PR #44, opened 2020-06-08
Fix watcher

dmolina79:
https://app.clubhouse.io/terminalsystems/story/16279/change-file-service-layer

Change Log:
* refactors Watcher into interface
* implements adding files on openfile to watcher
* removes logic for watching app folder by default
* fixes unit tests

TODO: finish the card on boot when we sync with @perfectmak


core/watcher/watcher.go
```diff
@@ -33,22 +40,12 @@ type FolderWatcher struct {
...
 // New creates an new instance of folder watcher
-// It listens to the path specified in the config space/folderPath
-func New(configs ...Option) (*FolderWatcher, error) {
+func New(configs ...Option) (FolderWatcher, error) {
 	options := watcherOptions{}
 	for _, config := range configs {
 		config(&options)
 	}
 
-	if len(options.paths) == 0 {
```
dmolina79, line 43: no longer watching home folder by default

PR #48, opened 2020-06-10
Installer Script

dmerrill6:
# Change log

- Refactor config/envs so that Textile Hub url is now part of space.json, and removed space folder from space.json (now we would be using storePath for both db and temp files).

- Add installer script as a new cmd.

- Add installer to goreleaser.


cmd/installer/main.go
```diff
@@ -0,0 +1,161 @@
...
+	"net/http"
+	"os"
+	"runtime"
+	"strings"
+
+	"github.com/dustin/go-humanize"
+)
+
+const (
+	tag            = "v0.0.1"
```
dmolina79, line 18: nice, this could potentially be captured from github envs later

grpc/pb/space.proto
```diff
@@ -37,9 +37,8 @@ message ListDirectoriesResponse {
 }
 
 message ConfigInfoResponse {
-  string folderPath = 1;
-  string port = 2;
-  string appPath = 3;
+  string port = 1;
```
dmolina79, line 40: need to coordinate with FE to sync on proto schema, since this could break the release on their side

.goreleaser.yml
```diff
@@ -13,6 +13,13 @@ builds:
   goos:
     - darwin
     - linux
+- env:
```
dmolina79, line 16: cool, we can test if it supports two binaries, im guessing it does

PR #49, opened 2020-06-11
Watcher and textile refactor

dmolina79:
https://app.clubhouse.io/terminalsystems/story/16279/change-file-service-layer
Change Log: (see write up for details)
* refactor watcher to sync with open files
* major refactor to textile package to handle per bucket concurrency
*  update code and tests to use new types and interface

Main
Screen Shot 2020-06-11 at 1 25 49 PM

Motivation for refactoring:
While finishing the watcher integration with syncing open files I realized that we need to handle the concurrency at the bucket level since several go-routines may be doing operations concurrently with a bucket and need to keep consistency on those operations, so that means we need to lock on write operations per bucket and the current design for the textile client had difficulty supporting that locking mechanism.

With this new design each bucket handles a lock internally on write and read operations


core/textile/bucket.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"github.com/ipfs/interface-go-ipfs-core/path"
+	"github.com/textileio/go-threads/core/thread"
+	bucketsClient "github.com/textileio/textile/api/buckets/client"
+	bucketsproto "github.com/textileio/textile/api/buckets/pb"
+	"io"
+	"sync"
+)
+
+type BucketsClient interface {
```
dmolina79, line 13: interface for textiles buckets client

PR #50, opened 2020-06-12
Move to daemon subfolder and remove poc

dmerrill6:
After this is merged, will rename the repo to `space`


.goreleaser.yml
```diff
@@ -6,18 +6,18 @@ before:
...
+    main: ./daemon/cmd/space/main.go
+    binary: space
     goos:
       - darwin
       - linux
   - id: installer
     env:
       - CGO_ENABLED=0
-    main: ./cmd/installer/main.go
+    main: ./daemon/cmd/installer/main.go
```
dmolina79, line 20: i am wondering about adding daemon, does this mean the intention is for this to become a monorepo?

dmerrill6: I'm adding another folder for server-side stuff (address book, vault, etc). Maybe later we can add the FE and bundle everything in one github actions flow

dmerrill6:
> LGTM. Was wondering if we should spaced as the repo name to distinguish it from the FE or other apps we may build.

spaced as in space daemon? Maybe we can keep this repo for the daemon but that would mean another repo for server side stuff. What do you think is best?

dmerrill6:
Closed in favor of multi-repo approach for now

PR #52, opened 2020-06-15
add progress uploads logic

dmolina79:
https://app.clubhouse.io/terminalsystems/story/16570/upload-file-send-back-to-the-ui-the-amount-of-bytes-remaining

Change Log:
* adds bytes and file and folder count to stream addItems
* updates proto schema
* updates tests
Screen Shot 2020-06-15 at 2 03 52 PM


core/space/services/services_fs.go
```diff
@@ -156,29 +156,125 @@ func (s *Space) createFolder(ctx context.Context, path string, b textile.Bucket)
...
+				if err != nil {
+					log.Error(fmt.Sprintf("error reading folder path %s ", pathInFs), err)
+					filesRes <- domain.AddItemsResponse{
+						Error: err,
+					}
+					return
+				}
+				for _, file := range files {
+					subPath := pathInFs + "/" + file.Name()
+					if subPath != pathInFs {
```
jsonsivar, line 212: Was having a little trouble understanding this piece.  This means when the contents are a file `subPath == pathInFs`?

dmolina79: yeah, not sure its needed honestly, I think i had an issue when using `Walk` method that needed that check, this is a precaution to avoid adding the existing path to the subpaths and dont get into an infinite recursive loop.

PR #54, opened 2020-06-17
Feature: Mounting Fuse Drive on startup and through gRPC

perfectmak:
**CHANGELOG**
- Update app initialization to conditionally mount a fuse drive if previously mounted or environment set to initialize
- Update grpc protobuf and server to allow mounting through grpc
- Implements SpaceFSDataSource that uses the Space Service to provide data for Fuse VFS.

There is still a bunch of code/tests upcoming for editing files, but though this is a good point to open a PR for this stage and subsequent PR will be added. But with this, we can boot up the daemon and have the drive mounted. Also navigating folders and files in the space.


app/app.go
```diff
@@ -8,6 +8,14 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/pkg/errors"
+
+	fuse2 "github.com/FleekHQ/space-poc/core/space/fuse"
+
+	"github.com/FleekHQ/space-poc/core/fs_data_source"
+
+	fuse "github.com/FleekHQ/space-poc/core/libfuse"
```
dmolina79, line 17: first could be libfuse and our space/fuse becomes fuse to avoid confusion

app/app.go
```diff
@@ -80,10 +88,22 @@ func Start(ctx context.Context, cfg config.Config, env env.SpaceEnv) {
 		space.WithAddWatchFileFunc(sync.AddFileWatch),
 	)
 
+	// setup FUSE FS Handler
+	fuseHandler, err := spacefs.NewSpaceFS(ctx, fs_data_source.NewSpaceFSDataSource(sv))
```
dmolina79, line 92: `spacefs.New()` should be fine to without repeating in the factory method

app/app.go
```diff
@@ -8,6 +8,14 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/pkg/errors"
+
+	fuse2 "github.com/FleekHQ/space-poc/core/space/fuse"
+
+	"github.com/FleekHQ/space-poc/core/fs_data_source"
```
dmolina79, line 15: convention here is to be names something like `fsds` in a single word like `strconv`

app/app.go
```diff
@@ -80,10 +88,22 @@ func Start(ctx context.Context, cfg config.Config, env env.SpaceEnv) {
 		space.WithAddWatchFileFunc(sync.AddFileWatch),
 	)
 
+	// setup FUSE FS Handler
+	fuseHandler, err := spacefs.NewSpaceFS(ctx, fs_data_source.NewSpaceFSDataSource(sv))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	mountPath := cfg.GetString(config.FuseMountPath, "~/")
```
dmolina79, line 97: cant we just hide away this bootstrap piece within a Start method thru the fuse interface we provide? seems that app.go is getting to big and we just need the `fuseController` at this level so a `New` method that sets up everything to return a fuseController will be cleaner from the app.go

app/app.go
```diff
@@ -126,6 +153,13 @@ func Start(ctx context.Context, cfg config.Config, env env.SpaceEnv) {
 		sync.Stop()
 	}
 
+	if fuseController != nil {
```
dmolina79, line 156: if fuseController can be nil in some path, it means line 116 can blow up?

core/fs_data_source/file_handler.go
```diff
@@ -1,34 +1,37 @@
...
+type ReadWriteSeekerCloser interface {
+	io.Closer
+	io.ReadWriteSeeker
 }
 
-func NewIPFSReadHandler(reader io.ReadSeekCloser) *IPFSFileHandler {
-	return &IPFSFileHandler{
-		isRead: true,
-		reader: reader,
+type FileHandler struct {
```
dmolina79, line 10: FileHandler can be private and in the New return and interface?

app/app.go
perfectmak: Oh true, I missed this part. The reason I created FuseController is to move most of these initializations away from app.go because it was becoming harder to understand what was going on as per initialization in that file. I'll work on moving this to FuseController also.

grpc/handlers.go
```diff
@@ -38,7 +40,7 @@ func (srv *grpcServer) SendTextileEvent(event events.TextileEvent) {
 }
 
 func (srv *grpcServer) ListDirectories(ctx context.Context, request *pb.ListDirectoriesRequest) (*pb.ListDirectoriesResponse, error) {
-	entries, err := srv.sv.ListDir(ctx)
+	entries, err := srv.sv.ListDir(ctx, "/")
```
dmolina79, line 43: @dmerrill6 is this the default path ^^^?

app/app.go
perfectmak: Ideally, it can't be null, I think I copied something else and just modify for this cleanup 😅. I'll fix it

grpc/handlers.go
perfectmak: Actually, the implementation for `ListDir()` in space Service does not work with that path. I was meaning to ask you how the frontend traverses the directories, if we are not passing a path to the `ListDirectories` grpc call? @dmolina79

dmolina79: ah ok then this is fine, i guess it could be empty `""` to mean the root of your space folder path, @dmerrill6 can confirm which one is better

core/space/fuse/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := s.store.Set([]byte(config.MountFuseDrive), []byte("true")); err != nil {
+		return err
+	}
+
+	s.serve()
+	return nil
+}
+
+func (s *Controller) serve() {
+	if s.isServed {
```
dmolina79, line 73: does this need a lock or are we always calling it from one go routine?

core/space/fuse/controller.go
```diff
@@ -0,0 +1,106 @@
...
+		log.Printf("Error fetching mountFuseDrive state: %s\n", err.Error())
+	}
+
+	return s.cfg.GetString(config.MountFuseDrive, "false") == "true"
+}
+
+// Mount mounts the vfs drive and immediately serves the handler.
+// It starts the Fuse Server in the background
+func (s *Controller) Mount() error {
+	if s.vfs == nil {
```
dmolina79, line 49: since this is called from grpc does it need to account concurrent access? is the internal vfs mount state safe to be accessed concurrently?

core/space/fuse/controller.go
perfectmak: No, I don't think so. It is currently only called from the `Mount()` function and that now has a lock

core/space/fuse/controller.go
perfectmak: No, the vfs implementation does not have a lock on Mounts, because the Mount function allows mounting multiple drives, so it could become complicated tracking and locking each drive name and mount path, so we leave it to error out from the operating systems FUSE server.

But I've updated the controller code add some locks, since the controller is currently only allowing a single mount call. We can make changes easily from here if we need to do more mounts or etc

core/fsds/spacefs.go
```diff
@@ -0,0 +1,102 @@
...
+	for i, domainEntries := range domainEntries {
+		dirEntries[i] = NewDirEntry(domainEntries.DirEntry)
+	}
+
+	return dirEntries, nil
+}
+
+// Open is invoked to read the content of a file
+func (d *SpaceFSDataSource) Open(ctx context.Context, path string) (ReadSeekCloser, error) {
+	bucketSlug := "" // bucketSlug is currently ignored, so defaulting this
```
dmerrill6, line 81: Maybe it makes sense to call textile client for list buckets and pick the first one, just so that when we add multiple buckets support it's easier to see where we need to refactor.

grpc/handlers.go
dmerrill6: This is a bit confusing, not sure why you are modifying the `ListDir` signature if path not going to be used anyways. Currently the FE lists the whole directory structure (that's what `ListDir` is returning). There's also the `listDirAtPath` method that you might want to make public if you need to list a specific path. But yea default value to list root and all subdirs should be `""`.

perfectmak:
> This is a bit confusing, not sure why you are modifying the ListDir signature if path not going to be used anyways. Currently the FE lists the whole directory structure (that's what ListDir is returning). There's also the listDirAtPath method that you might want to make public if you need to list a specific path. But yea default value to list root and all subdirs should be "".

@dmerrill6 I just realized the actual behaviour of that function after discussing with Doug. I might take your advice and expose the `ListDirAtPath` as public, because it is currently wasteful for the native FileSystem to list all the nested directory tree on a single call. After taking a closer look at the implementation, I'll decide which one is more efficient and use it.

grpc/handlers.go
perfectmak: @dmerrill6 I just realized the actual behaviour of that function after discussing with Doug. I might take your advice and expose the `ListDirAtPath` as public, because it is currently wasteful for the native FileSystem to list all the nested directory tree on a single call. After taking a closer look at the implementation, I'll decide which one is more efficient and use it.

PR #57, opened 2020-06-18
Add sharing endpoints to proto

dmerrill6:
Added sharing and user identity creation endpoints to proto.

Missing ACL modification endpoints.

[ch16553]


grpc/pb/space.proto
```diff
@@ -95,4 +106,66 @@ message CreateFolderRequest {
...
+message ShareBucketViaEmailRequest {
+  string bucket = 1;
+  string email = 2;
+}
+
+message ShareBucketViaEmailResponse {}
+
+enum IdentityType {
+  USERNAME = 0;
+  EMAIL = 1;
```
jsonsivar, line 153: Does email take precedence if they have both?

giankotarola:
@jsonsivar qq related to:
> Login or user initialization (either via private key, password or email depending on what we finalize)
>
 the password that you meant is related to this, right? Also, the password is an account password or a passphrase to encrypt the user key?
Screen Shot 2020-06-19 at 09 49 37

Another doubt is the username field validation. For validation are we going to have another endpoint or will be validate on the signup endpoint?

grpc/pb/space.proto
```diff
@@ -14,12 +14,23 @@ service SpaceApi {
   rpc OpenFile(OpenFileRequest) returns (OpenFileResponse) {}
   rpc AddItems(AddItemsRequest) returns (stream AddItemsResponse) {}
   rpc CreateFolder(CreateFolderRequest) returns (CreateFolderResponse) {}
-}
 
-message ListDirectoriesRequest {
+  // Identities
+  rpc GetIdentityByUsername(GetIdentityByUsernameRequest) returns (GetIdentityByUsernameResponse) {}
+  rpc CreateUsernameAndEmail(CreateUsernameAndEmailRequest) returns (CreateUsernameAndEmailResponse) {}
```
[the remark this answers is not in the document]
dmerrill6: This endpoint would encapsulate two backend calls, one to create a username and one to create an email for that username. If email is not provided it will only do the first call

grpc/pb/space.proto
```diff
@@ -95,4 +106,66 @@ message CreateFolderRequest {
...
 // not sure we need to return anything other than an error if we failed
 message CreateFolderResponse {
+}
+
+message GetIdentityByUsernameRequest {
+  string username = 1;
+}
+
+message Identity {
+  string address = 1;
```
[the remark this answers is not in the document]
dmerrill6: It's another type of address based on the public key (just like a crypto address in blockchain). This is just in case you want to send a message to a given public key and you don't have the username.

The address book models look like this:

Identity Model: address (pub key representation), username
Proof model: proof type (e.g. email, twitter handle), username

Identity has many proofs.

grpc/pb/space.proto
dmerrill6: This is just for the `ShareBucketViaIdentityRequest` method. So you can use this to either share a bucket to a user through username or through email (e.g. if I can send `ShareBucketViaIdentityRequest(USERNAME, "jsonsivar", "my-bucket")`). I guess we can add ADDRESS as an enum field too if we need it.

dmerrill6: > LGTM just had some clarifications.
> 
> Here are some other interfaces that come to mind. If you think they make sense, I can create a PR targeting your branch lmk:
> 
> Login or user initialization (either via private key, password or email depending on what we finalize)

That's partially handled by `RecoverKeysByPassphrase`. We can add another one to import private key too I think.

> Logout or unset user

Makes sense. But I'd call this `DeleteUserKeys` or something like that because that's what would be actually happening.

> Create team (based on a list of identities - maybe we can leave an optional field for read vs write which we can use later)

Makes sense, not too sure how that's handled by the daemon yet tho.

> Create shared bucket(bucket name, team id)

Or maybe just create bucket with a param to set if you want it shared or not. Maybe it can be part of the UI (e.g. there's a toggle next to each bucket to choose on/off for "automatic back up")

> Get team members (for now to use our own db and can upgrade to Textile when this gets added)

Sounds good

> Remove member (team id, identity)

Sounds good too. Maybe Add Member too?

dmerrill6:
> the password that you meant is related to this, right? Also, the password is an account password or a passphrase to encrypt the user key?

Yes the password is the one for the image you sent. This would be sent to the `BackupKeysByPassphrase` rpc. There's no account password, only passphrase to encrypt the user key (on that note I'd maybe make some changes in the UI to be more clear about this, adding a) a warning about that password not being recoverable and b) triggering the system keystore to ask the users if they want to back up their password in their key chain)

> Another doubt is the username field validation. For validation are we going to have another endpoint or will be validated on the signup endpoint?

That's the `GetIdentityByUsername` endpoint. If it returns anything it means the username is not available. If it returns an error then the username is available. Although if you try to send an already used username to the `CreateUsernameAndEmail` it will throw an error too.

giankotarola:
> > the password that you meant is related to this, right? Also, the password is an account password or a passphrase to encrypt the user key?
> 
> Yes the password is the one for the image you sent. This would be sent to the `BackupKeysByPassphrase` rpc. There's no account password, only passphrase to encrypt the user key (on that note I'd maybe make some changes in the UI to be more clear about this, adding a) a warning about that password not being recoverable and b) triggering the system keystore to ask the users if they want to back up their password in their key chain)
> 
Ok, but I remembered that yesterday I asked @Blockchainbrett about it and mentioned that should be a password account instead of a passphrase to encrypt the user key, so I'm a little confused. Also, if we go in that way, I think it would be better to change some text on the "Password" modal, so users know that the password is for encrypt their api key and it is not a password account.
> > Another doubt is the username field validation. For validation are we going to have another endpoint or will be validated on the signup endpoint?
> 
> That's the `GetIdentityByUsername` endpoint. If it returns anything it means the username is not available. If it returns an error then the username is available. Although if you try to send an already used username to the `CreateUsernameAndEmail` it will throw an error too.
>
If `GetIdentityByUsername` return something should not mean that the username is available and if return an error should not mean the user is not available?

dmerrill6:
> Ok, but I remembered that yesterday I asked @Blockchainbrett about it and mentioned that should be a password account instead of a passphrase to encrypt the user key, so I'm a little confused. Also, if we go in that way, I think it would be better to change the texts on the "Password" modal, so users know that the password is for encrypt their api key and it is not a password account.

There's no such thing as accounts 🤔 But the functionality should be the same that what the designs show. It's a password that encrypts the private key. I guess it's fine if the user thinks it's an account password as long as they know that we are not storing it and only by using that password they can recover their keys if they log out.

> If `GetIdentityByUsername` return something should not mean that the username is available and if return an error should not mean that the user is not available?

If `GetIdentityByUsername` returns something, it means that there's an identity using that username so it means the username is taken.

Blockchainbrett:
Yea, this password would basically be able to login later on. They can log in to the same device or different device and it would import their keys. Whether we call it an "account" or not, that functionality is good. 

1) they don't set the password at login
2) they set the password if they want to logout, maybe to make a second account on the same device
3) they log back in on the same device with that username and password
4) they can log into a different device with that username and password

PR #58, opened 2020-06-21
possible fix for high CPU usage

dmolina79:
NOTE: after testing with the profile analysis, i think we need to handle when the channel is closed by threads library to avoid being stuck on those go routines in an infinite loop. 

We still need to monitor the app and CPU is something else pops up, since its hard to replicate given that the usually could be related to threads side.

### DIFF
--- a/core/textile/listener.go
+++ b/core/textile/listener.go
@@ -57,6 +57,10 @@ func (tl *textileThreadListener) Listen(ctx context.Context) error {
 	}
 
 	channel, err := threads.Listen(bucketCtx, *dbID, []threadsc.ListenOption{opt})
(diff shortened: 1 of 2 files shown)


core/textile/listener.go
```diff
@@ -87,6 +92,9 @@ func (tl *textileThreadListener) Listen(ctx context.Context) error {
 			case val, ok := <-channel:
 				if ok {
 					listenerEventHandler(val)
+				} else {
+					tl.Close()
```
dmolina79, line 96: if the channel is closed, there is no point on keeping the loop alive, we must consider logic around restarting listeners if this happen

PR #64, opened 2020-06-24
add default values config

dmolina79:
https://app.clubhouse.io/terminalsystems/story/16673/run-without-configuration-env-vars

change Log:
* Adds map backed implementation for config with default values
* Adds default env implementation
* removes JSON config file dependency from daemon and installer
* deprecated Implementation for Json backed config

TODO: move out of .env file by writing textile keys to store in installer and implement env that source from store

### DIFF
--- a/config/map_config.go
+++ b/config/map_config.go
@@ -1,0 +1,55 @@
+package config
+
+import (
+	"github.com/FleekHQ/space-poc/core/env"
+)
+
+type mapConfig struct {
+	configStr map[string]string
+	configInt map[string]int
+}
+
+func NewMap(env env.SpaceEnv) Config {
+	configStr := make(map[string]string)
+	configInt := make(map[string]int)
+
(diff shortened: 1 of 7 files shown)


config/map_config.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	if stringValue, ok := defaultValue.(string); ok {
+		return stringValue
+	}
+
+	return ""
+}
+
+func (m mapConfig) GetInt(key string, defaultValue interface{}) int {
```
perfectmak, line 45: why don't we make the `defaultValue` an `int` here since we are expecting an int to be returned. Same for the `GetString()`'s default value

dmolina79: Good point, i was mimicking the json config library method signature i was using before, but i think there is no problem in using int here, will change and try to compile

PR #70, opened 2020-06-29
Feature/ch16524/create bucket grpc

jsonsivar:
### Change Log
* Create bucket grpc endpoint
* Join thread test (to be used in next card)


grpc/handlers_textile.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func (srv *grpcServer) CreateBucket(ctx context.Context, request *pb.CreateBucketRequest) (*pb.CreateBucketResponse, error) {
+	b, err := srv.sv.CreateBucket(ctx, request.Slug)
+	if err != nil {
+		return nil, err
+	}
+
+	br := &pb.Bucket{
+		Key:       b.GetData().Key,
```
perfectmak, line 16: Why not call `b.GetData()` once and use the returned value to populate pb.Bucket here

jsonsivar: Fixed in [bcb5c43](https://github.com/FleekHQ/space-poc/pull/70/commits/bcb5c43795de8e5bc9d092325fd651b03ff8e00d).

PR #71, opened 2020-06-30
REST API Proxy

dmolina79:
**CHANGELOG**
* updates to proto file to support REST annotations
* moves proto file into a better structure inside grpc folder
* fixes package generation issue of PB go code to be in correct package.
* adds makefile commands to make generation of pb and proxy easier.
* Adds Support for grpc reverse REST proxy. Starts on port 9997 by default
* Update definition for REST endpoints in space.proto file. Also, add comments that help with swagger docs
* Embedded swagger ui in daemon, for easy testing and access to the REST API.

After running the space app, visit http://localhost:9997/swaggerui/ to see and test the rest endpoints.

Screenshot 2020-07-01 at 22 17 01

References:
https://github.com/grpc-ecosystem/grpc-gateway
https://www.youtube.com/watch?v=Zf9G2KzYs7w


grpc/proto/space.proto
```diff
@@ -1,17 +1,30 @@
...
+import "google/api/annotations.proto";
 
 package space;
 
+option go_package = ".;pb";
+
 service SpaceApi {
   rpc ListDirectories(ListDirectoriesRequest) returns (ListDirectoriesResponse) {}
-  rpc GetConfigInfo(google.protobuf.Empty) returns (ConfigInfoResponse) {}
+  rpc GetConfigInfo(google.protobuf.Empty) returns (ConfigInfoResponse) {
```
dmolina79, line 13: Need to define all rest endpoints how they will look like.

grpc/proto/space.proto
```diff
@@ -1,17 +1,30 @@
 syntax = "proto3";
 
+
 import "google/protobuf/empty.proto";
+import "google/api/annotations.proto";
 
 package space;
 
+option go_package = ".;pb";
```
dmolina79, line 9: this sets the generated file to the correct package on the output.

Makefile
```diff
@@ -12,4 +12,14 @@ test:
 	go test ./...
 
 proto_gen:
-	protoc -I grpc/pb/ grpc/pb/space.proto --go_out=plugins=grpc:grpc/pb
\ No newline at end of file
+	protoc -I grpc/pb/ -I grpc/proto/ -I${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis grpc/proto/space.proto --go_out=plugins=grpc:grpc/pb
+
+gen_proxy:
```
jsonsivar, line 17: Maybe we call this `gen_gateway` or `gen_rest` since it might get confused with the web proxy.

grpc/grpc.go
```diff
@@ -63,21 +73,84 @@ func New(sv space.Service, fc *fuse.Controller, opts ...ServerOption) *grpcServe
...
+			mux.ServeHTTP(w, r)
+		}
+	}))
+
+	go func() {
+		if err := srv.restServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Error(fmt.Sprintf("REST server failed to start on port %d", srv.opts.restProxyPort), err)
+		}
+	}()
+	log.Info(fmt.Sprintf("REST server started on Port %v", srv.opts.restProxyPort))
```
jsonsivar, line 137: Wouldn't this log even if the server failed to start since `ListenAndServe` is in a go routine?

grpc/grpc.go
```diff
@@ -129,6 +198,15 @@ func WithProxyPort(port int) ServerOption {
 	}
 }
 
+// WithRestProxyPort configures the REST Proxy port
+func WithRestProxyPort(port int) ServerOption {
+	return func(o *serverOptions) {
```
jsonsivar, line 203: This is more for knowledge but why return a func that sets it instead of just having the `(o *serverOptions)` in the line above?

perfectmak: This is generally a pattern for exposing configurations parameters in go. Admittedly, this particular option doesn't do much, but there are cases where the config does some more complex checks and uses some internal values that is doesn't want to expose to the code using the module.

grpc/grpc.go
perfectmak: Yeah, true, but if `ListenAndServe` fails we get another logged error message.
Just to make it more clearing I'm updated this log message to `REST server is starting`

Makefile
perfectmak: Fixed

PR #74, opened 2020-07-02
Integrate with Textile Hub Lambda

dmerrill6:
# Change log

- Replaces authentication to Textile hub so that keys aren't stored in the client side.

- Connects to a lambda through web sockets to get the auth token.

[ch16547]


config/map_config.go
```diff
@@ -19,7 +19,8 @@ func NewMap(envVal env.SpaceEnv, flags *Flags) Config {
 	configStr[SpaceStorePath] = "~/.fleek-space"
 	configStr[TextileHubTarget] = "textile-hub-dev.fleek.co:3006"
 	configStr[TextileThreadsTarget] = "textile-hub-dev.fleek.co:3006"
-	configStr[SpaceServicesAPIURL] = "https://td4uiovozc.execute-api.us-west-2.amazonaws.com/dev" // TODO: Get a domain
+	configStr[SpaceServicesAPIURL] = "https://td4uiovozc.execute-api.us-west-2.amazonaws.com/dev"   // TODO: Get a domain
```
perfectmak, line 22: we should probably pass this values through the build command/flags like the mongo information shown below.
So it is easily configurable on build

core/textile/client.go
```diff
@@ -102,41 +100,18 @@ func (tc *textileClient) requiresRunning() error {
...
-	}
-
-	// TODO: CTX has to be made from session key received from lambda
-	log.Debug("Creating libp2p identity")
-	tok, err := tc.threads.GetToken(ctx, thread.NewLibp2pIdentity(privateKey))
+	log.Debug("Authenticating with Textile Hub")
+	tokStr, err := getHubToken(tc.store, tc.cfg)
 	if err != nil {
-		return nil, err
+		log.Error("Token Challenge Error:", err)
```
perfectmak, line 112: Shouldn't you return the err here, or is it possible to initialize `thread.Token()` with an empty `tokStr`?

config/map_config.go
jsonsivar: Yeah this sounds good, maybe for the Textile Hub targets as well.

core/textile/client.go
jsonsivar: Agreed

dmerrill6: Fixed in 4ef068e

config/map_config.go
dmerrill6: Fixed in 4ef068e

PR #76, opened 2020-07-02
Feature: Implement ListDirectory

perfectmak:
**CHANGELOG**
- Add ListDirectory gRPC method that returns only items of a directory and not is subdirectories
- Fix cpu profiling not dumping results on exits

** TODO **
- Figure out why multiple grpc calls errors with too many open files when interacting with textile bucket. Currently following this lead https://godoc.org/google.golang.org/grpc#ClientConn.NewStream and profiling


perfectmak:
I'm going to rename this to just an implementation of ListDirectory and merge, so we can quickly rename the repository and directories.

PR #78, opened 2020-07-03
Proto with list buckets and bucket param

dmerrill6:
# Change log

- Update the proto to include `ListBuckets` method.

- Update the proto to include the param `bucket` in methods that require (we are currently using default bucket, but moving away from that).

## TODO

Will update the pb once this is approved and create stubs / update the methods that require bucket name once the proto is approved.

### DIFF
--- a/grpc/proto/space.proto
+++ b/grpc/proto/space.proto
@@ -167,9 +167,17 @@ service SpaceApi {
       body: "*"
     };
   }
+
+  rpc ListBuckets(ListBucketsRequest) returns (ListBucketsResponse) {
+    option (google.api.http) = {
+      get "/v1/buckets"
+    }
+  }
 }
 
-message ListDirectoriesRequest {}
+message ListDirectoriesRequest {
+  string bucket = 1;
+}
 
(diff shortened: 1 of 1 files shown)


dmerrill6:
> @dmerrill6 LGTM. is this the final version for the client release?

It should be, unless I'm missing anything, @jsonsivar @perfectmak please confirm

grpc/proto/space.proto
```diff
@@ -167,9 +167,17 @@ service SpaceApi {
...
+  rpc ListBuckets(ListBucketsRequest) returns (ListBucketsResponse) {
+    option (google.api.http) = {
+      get "/v1/buckets"
+    }
+  }
 }
 
-message ListDirectoriesRequest {}
+message ListDirectoriesRequest {
+  string bucket = 1;
```
jsonsivar, line 179: Okay so based on the discussion in Slack, I think we need `path` here as well. @dmerrill6 could you add it and regen it? Then I will use that as a base to fix the listdir(s) issues.

dmerrill6:
Already merged in #79

PR #79, opened 2020-07-03
Bucket name as param

dmerrill6:
# Change log

- Update textile operations to receive a `bucket` param so that they work for buckets beyond the default one.

- Update generators in makefile to use a local reference so in that way avoiding dependencies on global Go packages for code generation (@perfectmak suggestion). Uses git submodules for these dependencies.

- Add documentation in Readme about using submodules.

- Includes proto updates in #78 and adds a stub for ListBuckets method

## TODO

Support non default buckets in FUSE (need @perfectmak 's help for this, defaulting the bucket for now and left a TODO in the code).

[ch16977]


README.md
```diff
@@ -134,11 +134,23 @@ For Linux it needs to be built from source.
...
 
+If you don't have swagger generator, you will need it for generating the REST API documentation:
+
+`brew install swagger-codegen`
+`brew install statik`
+
+Finally, you will need to install the git submodules for some extra dependencies in the proto buffer generation:
+
+```
+git submodule init
```
perfectmak, line 150: I was just thinking we copied the `googleapi` directly into our codebase, so we can avoid using submodules or depending on their repository. This way we can be sure that they don't introduce any breaking changes we are not aware of (although, I don't think the protocol buffer definitions will be changing that often) and also no need for submodule inits. Let me know what you think

core/fsds/spacefs.go
```diff
@@ -52,7 +52,9 @@ func (d *SpaceFSDataSource) Get(ctx context.Context, path string) (*DirEntry, er
 
 	baseName := filepath.Base(path)
 	parentPath := filepath.Dir(strings.TrimRight(path, "/") + "/..")
-	parentEntries, err := d.service.ListDir(ctx, parentPath)
+
+	// TODO: Support non default bucket, passing "" for now which is default bucket
+	parentEntries, err := d.service.ListDir(ctx, parentPath, "")
```
perfectmak, line 57: Yeah, with this ListBuckets approach, the FUSE fs logic would need some changes to account for the list of buckets are the root directory and then browse the files on subsequent subdirectories. I'll create a card to track this

README.md
dmerrill6: Fixed in 4855f844713150c70853c9596be8494d139a4485

PR #80, opened 2020-07-07
Hotfix: Mongo connection error(cannot unmarshall DNS)

jsonsivar:
### Change Log
* Use updated MongoDB connection string to prevent related bug (https://stackoverflow.com/questions/55660134/cant-connect-to-mongo-cloud-mongodb-database-in-golang-on-ubuntu)
* Update dev and release configurations to include new environment variable
* Update git action files

Background: Some Linux distros were getting an error connecting to Mongo using the SRV endpoint.

Note: Mongo is temporarily being used for some bucket specific metadata which is soon to be replaced with an embedded data store.


jsonsivar:
> LGTM!. We just need to test it on macOS too.

Agreed, can someone with macOS here volunteer to  run this and make sure it still works? Please note that one env var will be updated and one env var is new (ping me for the values).

dmerrill6:
> > LGTM!. We just need to test it on macOS too.
> 
> Agreed, can someone with macOS here volunteer to run this and make sure it still works? Please note that one env var will be updated and one env var is new (ping me for the values).

I can test it now

dmerrill6:
> > > LGTM!. We just need to test it on macOS too.
> > 
> > 
> > Agreed, can someone with macOS here volunteer to run this and make sure it still works? Please note that one env var will be updated and one env var is new (ping me for the values).
> 
> I can test it now

Worked after deleting ~/.fleek-space 👍

PR #82, opened 2020-07-07
Fix: Limit ThreadsDB p2p connection

perfectmak:
This PR allows us to configure the maximum number of open TCP connections created by textiles threads p2p connections.
By default the Maximum connection is set to 100, which is low enough to not reach the file descriptor limits on devices we've tested (currently macOS appears to default to 256 file descriptor limit).

NOTE: This PR modifies the textile dependency to use the version containing the fix in my fork. There is a PR open for it [here](https://github.com/textileio/textile/pull/261) to merge it to their release.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,6 +22,8 @@ const (
 	Mongopw                  = "space/mongoPw"
 	Mongohost                = "space/mongoHost"
 	Mongorepset              = "space/mongoRepSet"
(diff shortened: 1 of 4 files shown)


jsonsivar:
Ah looks like the previous merge created some conflicts for this PR. If we handle those it should be good to merge.

perfectmak:
@jsonsivar conflict has been fixed now

PR #85, opened 2020-07-09
Bucket list and Textile Client refactor

dmerrill6:
# Change log

- Implement List Buckets gRPC

- Implement meta thread to keep bucket state in textile bucket

- Refactor and cleanup of Textile Client so that's easier to contribute and follow

- Added readme for documenting our Textile Client 

- Update Join and Share bucket to use meta thread instead of store

- Add vscode launch configuration to make debugging with breakpoints easier

[ch16836] [ch16525]


core/textile/bucket/bucket.go
```diff
@@ -21,30 +33,31 @@ type BucketsClient interface {
 // NOTE: all write operations should use the lock for the bucket to keep consistency
 // TODO: Maybe read operations dont need a lock, needs testing
 // struct for implementing bucket interface
-type bucket struct {
+type Bucket struct {
 	lock          sync.RWMutex
 	root          *bucketsproto.Root
-	client        Client
+	ctx           context.Context
```
perfectmak, line 39: It is usually an anti-pattern to store a context in a struct.

The context should be only valid for a particular period of operation, that is why it is usually passed into running functions to control their context values and trigger them to be done.

Structs objects could exists for a long time beyond the lifetime of a context and therefore all operations on that Bucket would stop executing.

dmerrill6: @perfectmak Added context back in bfc5cf6.

Please check it out and let me know if there's any other issues.

dmerrill6:
> Looks good to me. Let me know when this get's merged, so I can rebase my pending branch and handle most of the changes that affect my branch too.

@perfectmak it's merged!

PR #86, opened 2020-07-10
Add integration test location and initial test

jsonsivar:
Integration tests can be run with `make integration_test` (and should be successful if you have the required environment variables exported) and unit tests will only run with `make test`.


integration_tests/test_fixtures.go
```diff
@@ -0,0 +1,52 @@
+package integration_tests
+
+import (
+	"github.com/FleekHQ/space-daemon/config"
+	spaceEnv "github.com/FleekHQ/space-daemon/core/env"
+	"github.com/namsral/flag"
```
perfectmak, line 6: Not I'm using `https://github.com/namsral/flag` here instead of the std go `flag` module. They have similar interfaces, it's just that this flag module also checks the environment for keys similar to the uppercase version of flag names.

perfectmak:
Closing this in favour of https://github.com/FleekHQ/space-daemon/pull/268

PR #88, opened 2020-07-11
Bug/ch16389/reenable windows

jsonsivar:
This re-enables windows build in the go-releaser.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -65,9 +65,9 @@ builds:
     binary: space
     goos:
       - windows
-    ignore:
-      - goos: windows
-        goarch: 386
+    # ignore:
+    #   - goos: windows
+    #     goarch: 386
 archives:
 - name_template: '{{ .Binary }}_{{ .Os }}_{{ .Arch }}'
   format: binary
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -47,6 +46,28 @@ builds:
...
+      - -X main.spacehubauth={{ .Env.SERVICES_HUB_AUTH_URL }}
+      - -X main.textilehub={{ .Env.TXL_HUB_TARGET }}
+      - -X main.textilehubma={{ .Env.TXL_HUB_MA }}
+      - -X main.textilethreads={{ .Env.TXL_THREADS_TARGET }}
+    main: ./cmd/space-daemon/main.go
+    binary: space
+    goos:
+      - windows
+    ignore:
+      - goos: windows
```
jsonsivar, line 69: @perfectmak @dmerrill6 I noticed in earlier versions 386 was disabled for windows. Do you guys know the reason for that? Kept it disabled in case it breaks the build but and we can enable it again and for now just enabled x86 on windows.

jsonsivar: Actually, I tested it locally and the goreleaser build successfully for both architectures so reenabled it.

PR #91, opened 2020-07-13
Feature: GenerateFileSharingLink

perfectmak:
**CHANGELOG**
- Encrypts file from Buckets with a new temp key
- Uploads encrypted file to ipfs
- OpenPublicSharedFile rpc call to decrypt and return its file location for opening

Sample Response via the rest interface (uploaded our LICENSE file):
```
{
  "link": "https://space.storage/files/share?fname=LICENSE&hash=Qmb4xixzW4s6b7LjL3AhLrg36Ndepjjam6mB3gpLtQwFG4&key=v571gWvymBWRkaZVN%2BTWcJnTo5NiK7%2BRBSasqRJQ2nKD4ln69XmORsV%2BFQqr%2FIEL",
  "fileCid": "Qmb4xixzW4s6b7LjL3AhLrg36Ndepjjam6mB3gpLtQwFG4",
  "fileKey": "v571gWvymBWRkaZVN+TWcJnTo5NiK7+RBSasqRJQ2nKD4ln69XmORsV+FQqr/IEL"
}
```

Screenshot 2020-07-17 at 14 19 21


core/space/services/services_sharing.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	// TODO: Publish file to ipfs
+
+	return &domain.FileSharingInfo{
+		Bucket:         bucketName,
+		Path:           path,
+		SharedFileHash: "",
+		SharedFileKey:  writer.EncodeKey(),
+		Link:           "https://space.storage/shared/{hash}/{decryption_key}",
```
perfectmak, line 52: @jsonsivar @dmerrill6 What should be the format of the link we would be sharing? This is currently the format I have in mind. Ideally, the link should map to a server we have access to, so we can fetch the hash from ipfs and decrypt with the key provided.

jsonsivar: I was thinking if we upload the file along with an index.html both in a folder to IPFS, then any gateway can serve up the decrypted file without depending on a backend server.

perfectmak: Yeah, I'm doing some tests to know what the content of the index.html would be, so as to decrypt the file and display it.

jsonsivar: Nice! On that note I suppose the decryption key would be a query string parameter in the link here. 

Also regarding the publishing to IPFS, was looking at some examples so just pasting some snippet here:
```
import httpapi "github.com/ipfs/go-ipfs-http-client"
IpfsAddr = tb.cfg.GetString(config.Ipfsaddr, "/ip4/127.0.0.1/tcp/5001")

ic, err := httpapi.NewApi(conf.AddrIPFSAPI)
if err != nil {
  return nil, err
}

pth, err := ic.Unixfs().Add(
  ctx,
  ipfsfiles.NewReaderFile(r),
  options.Unixfs.CidVersion(1),
  options.Unixfs.Pin(true))
if err != nil {
  return nil, err
}

return ic.ResolveNode(ctx, pth)
```

PR #92, opened 2020-07-15
Feature/ch17045/save join meta

jsonsivar:
### Change Log
* Generate new "root" keypair for signing tokens on startup
* Refactor some keychain functions for reuse
* Sign and add token to context for all requests going forward
* Save share and join records in bucket meta thread including pubkey

### TODO
Tests were done using the space-client-example but will add unit tests as well


core/keychain/keychain.go
```diff
@@ -38,26 +44,43 @@ func New(store db.Store) *keychain {
...
 func (kc *keychain) GenerateKeyPair() ([]byte, []byte, error) {
-	if val, _ := kc.store.Get([]byte(PublicKeyStoreKey)); val != nil {
+	return kc.GenerateNamedKeyPair(PrivateKeyStoreKey, PublicKeyStoreKey)
+}
+
+func (kc *keychain) GenerateRootKeyPair() ([]byte, []byte, error) {
+	return kc.GenerateNamedKeyPair(RootPrivateKeyStoreKey, RootPublicKeyStoreKey)
+}
+
+func (kc *keychain) GenerateNamedKeyPair(privkeystorekey string, pubkeystorekey string) ([]byte, []byte, error) {
```
perfectmak, line 54: perhaps this should be a private function and not a public one? It looks more like an internal implementation details.

core/keychain/keychain.go
```diff
@@ -38,26 +44,43 @@ func New(store db.Store) *keychain {
...
 func (kc *keychain) GetStoredKeyPairInLibP2PFormat() (crypto.PrivKey, crypto.PubKey, error) {
+	return kc.GetStoredKeyPairByNameInLibP2PFormat(PrivateKeyStoreKey, PublicKeyStoreKey)
+}
+
+func (kc *keychain) GetRootKeyPairInLibP2PFormat() (crypto.PrivKey, crypto.PubKey, error) {
+	return kc.GetStoredKeyPairByNameInLibP2PFormat(RootPrivateKeyStoreKey, RootPublicKeyStoreKey)
+}
+
+// Returns the stored key pair using the same signature than libp2p's GenerateEd25519Key function
+func (kc *keychain) GetStoredKeyPairByNameInLibP2PFormat(privkeystorekey string, pubkeystorekey string) (crypto.PrivKey, crypto.PubKey, error) {
```
perfectmak, line 73: Same with this. It should be a private function also

core/textile/client.go
```diff
@@ -162,12 +162,20 @@ func (tc *textileClient) Start(ctx context.Context, cfg config.Config) error {
...
+	// space app keys
 	if _, _, err := kc.GenerateKeyPair(); err != nil {
 		log.Debug("Starting Textile Client: Error generating key pair, key might already exist")
 		log.Debug(err.Error())
 		// Not returning err since it can error if keys already exist
 	}
 
+	// root signing keys
+	if _, _, err := kc.GenerateRootKeyPair(); err != nil {
+		log.Debug("Starting Textile Client: Error generating key pair, key might already exist")
```
perfectmak, line 174: Maybe it would better if we actually check the error for an existing key error. So we can catch any other possible error that might have occurred and return that one.

core/textile/bucket_factory.go
```diff
@@ -259,3 +266,63 @@ func (tc *textileClient) JoinBucket(ctx context.Context, slug string, ti *domain
...
+
+	tok, _ := thread.TokenFromContext(ctx)
+	pk, err := tok.Validate(rk)
+	return pk, nil
+}
+
+func (tc *textileClient) saveInviteRecord(ctx context.Context, slug string) error {
+	bctx, dbID, err := tc.initBucketThreadMetaCollection(ctx, slug, bucketInviteCollectionName, &BucketInvite{})
+
+	pk, err := tc.getContextPubKey(ctx)
```
dmerrill6, line 285: Not sure if this is the public key we want, as this will be the public key of the current user and not the one that receives the invitation I think.

jsonsivar: The one that receives it is being saved on the `JoinThread` call.  I'm saving this one here as the sender pub key so maybe it's just a matter of calling the field `SenderPubKey` or something. I don't think our scoping called the need for this specifically but just saved it while I was doing the saves here.

core/textile/client.go
jsonsivar: Yeah I thought the same too looking at all the existing places we do this skipped to make this PR quick. I agree will use this PR to clean it up as it's in a few places.

jsonsivar:
> Not getting why we need a root key. Isn't this going to get messy when we export/import keys to a new device?

I was thinking the root key doesn't need to get exported/imported.  I just created that concept because for local signing we need another key (i.e., to do this https://github.com/textileio/go-threads/blob/master/core/thread/identity.go#L116). It felt weird using both the privkey and pubkey in that call using the same keypair so created the root key which is like an daemon key that signs local token requests.  I think this will be handy when we introduce other applications and we can have them pass through their key (or we generate one for them) to segregate access within a single instance of the daemon (like we discussed for gatekeeping different apps' data). 

Having said that open to suggestions if there is a better scheme for this.

jsonsivar:
Also closing this in favor of new PR.

PR #97, opened 2020-07-22
gRPC Refactor: Remove identity from grpc

dmerrill6:
# Change log

- Changed identity requiring methods to only use public keys.

- Removed methods for obtaining / creating identities. Those should now be called from apps implementing the Daemon.

- Renamed the gRPC that generates a public link for a file and changed the request shape to allow password and an array of files (@perfectmak).

- Renamed the SelectGroup method.

- Changed the returned bucket shape to only include public key of members instead of whole identity.

- Updated docs.

## TODO

Implement credentials for 3rd party app authorization into methods using https://grpc.io/docs/guides/auth/

Client will have to include credentials when calling the gRPC in the future. Think this approach is better than adding an additional param to each method called "token" and calling a validating function in each gRPC endpoint implementation.


grpc/proto/space.proto
```diff
@@ -27,13 +27,6 @@ service SpaceApi {
     };
   }
 
-  // Get Config information installed daemon
-  rpc GetConfigInfo(google.protobuf.Empty) returns (ConfigInfoResponse) {
```
perfectmak, line 31: Also, don't forget to delete the 
```
message ConfigInfoResponse {
  string port = 1;
  string appPath = 2;
}
```

dmerrill6: Fixed in 9de527b.
Will merge soon if we don't get additional comments.

PR #98, opened 2020-07-24
Feature/ch17238/multipe file sharing logic

perfectmak:
**CHANGELOG**
- Replace random encryption key with a pbkdf generation
- Update sharing logic to zip multiple files before encryption.
- Update open public shared filed logic to use password when decrypting.


perfectmak:
> LGTM. Wonder if we should move the pbkdf salt and IV to a config tho so that we inject it on build and not expose it in the repo. What do you think?

Yes, I can move it to a config, but I felt we should leave those values open because if we want people to build custom experience around publicly shared files without having to ask use for the decryption parameters. One update I want to do though is to have the IV be random and prepend it to the encrypted payload. So all that would be left is the salt and I think it is reasonably safe to leave that open. What do you think?

PR #101, opened 2020-07-27
GRPC update for invite msgs

jsonsivar:
The thinking here is that the Textile inboxing is behind the scenes so the GRPC methods just deal with invitations.  Was thinking of we put the invitation ID in the payload of the message, and also put the message ID in our invitation meta thread, then we can lookup them up 2 ways (e.g., when new message is received, we can look up the invitation ID for that message and send the FE the additional details). 

Here is the link to the main commit with hub inboxing from Textile: https://github.com/textileio/textile/commit/290e676889aec261d91d2fda5dcdd4aa1e60733d#diff-5055b2fe97b1d505b9ae516345fa1510R93.


core/events/events.go
```diff
@@ -42,3 +42,19 @@ func NewTextileEvent(bucketname string) TextileEvent {
...
+	Accepted InvitationStatus = "ACCEPTED"
+	Rejected InvitationStatus = "REJECTED"
+)
+
+type InvitationEvent struct {
+	InviterPublicKey string
+	InvitationID     string
+	Status           InvitationStatus
+	ItemPaths        []string
+	CreatedAt        int64
```
jsonsivar, line 58: Behind the scenes, our invitation metadata could have the Textile message ID and look up the created at and read at fields.

PR #103, opened 2020-07-28
Hotfix: misc timing issues

jsonsivar:
This should fix some timing related issues that are being seen on some machines (I think more so on smaller ones). It's not final because the right way to do it is to either get a channel from Textile or poll the port to see if its listening yet, but for now doing a `sleep` as a temp fix so hackathon users aren't blocked.

### Change Log
* Make `buckd` wait for ready signal before moving on the next init
* Disable listing buckets in the syncer (this is also causing issues so disabling it until we resolve it - assumed this was okay because the listener was already partly commented out)
* Check for errors in a few places we should have been that were swallowing up errors
* Use `localhost` instead of `127.0.0.1` for bucket client to avoid wierd ipv6/name resolution errors (I think)

### TODO
* If Textile doesn't have a channel or something we can use, poll for port instead of using a sleep
* Fix issue with syncer

### DIFF
--- a/core/textile/buckd.go
+++ b/core/textile/buckd.go
@@ -25,13 +25,13 @@ var MinThreadsConn int
 type TextileBuckd struct {
 	textile   *core.Textile
 	IsRunning bool
-	ready     chan bool
+	Ready     chan bool
 	cfg       config.Config
 }
 
 func NewBuckd(cfg config.Config) *TextileBuckd {
 	return &TextileBuckd{
-		ready: make(chan bool),
+		Ready: make(chan bool),
 		cfg:   cfg,
 	}
 }
@@ -105,13 +105,18 @@ func (tb *TextileBuckd) Start(ctx context.Context) error {
 	log.Info("Welcome to bucket", fmt.Sprintf("peerID:%s", textile.HostID().String()))
 	tb.textile = textile
 	tb.IsRunning = true
+	tb.Ready <- true
 	return nil
+}
+
+func (tb *TextileBuckd) WaitForReady() chan bool {
+	return tb.Ready
 }
 
 func (tb *TextileBuckd) Stop() error {
 	tb.IsRunning = false
 	tb.textile.Close()
-	close(tb.ready)
+	close(tb.Ready)
 	// TODO: what else
 	return nil
 }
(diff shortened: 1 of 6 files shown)


core/textile/buckd.go
```diff
@@ -103,15 +104,33 @@ func (tb *TextileBuckd) Start(ctx context.Context) error {
 	textile.Bootstrap()
 
 	log.Info("Welcome to bucket", fmt.Sprintf("peerID:%s", textile.HostID().String()))
+
+	// wait for the port to be listening to consider it done
+	log.Info("Waiting for 127.0.0.1 3006 to listen")
+	for {
```
perfectmak, line 110: Instead of a forever loop, you can replace this with a counted for loop so as to allow some sort of timeout and we can have a sleep() for a few ms between each iteration

core/textile/buckd.go
```diff
@@ -103,15 +104,33 @@ func (tb *TextileBuckd) Start(ctx context.Context) error {
...
+func (tb *TextileBuckd) WaitForReady() chan bool {
+	return tb.Ready
+}
+
 func (tb *TextileBuckd) Stop() error {
 	tb.IsRunning = false
 	tb.textile.Close()
-	close(tb.ready)
+	close(tb.Ready)
 	// TODO: what else
```
perfectmak, line 134: Maybe we can also remove this TODO 🙂

jsonsivar: Fixed.

core/textile/buckd.go
jsonsivar: This didn't work either way so went back to sleeping so at least it's not blocking users, and we can upgrade the check later.

core/textile/buckd.go
```diff
@@ -103,15 +104,33 @@ func (tb *TextileBuckd) Start(ctx context.Context) error {
 	textile.Bootstrap()
 
 	log.Info("Welcome to bucket", fmt.Sprintf("peerID:%s", textile.HostID().String()))
+
+	// wait for the port to be listening to consider it done
+	log.Info("Waiting for 127.0.0.1 3006 to listen")
```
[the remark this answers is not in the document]
jsonsivar: I changed it back to sleep, so thinking in that case it makes sense to keep it as info since it sleeps for 5s and good to inform the user.

PR #105, opened 2020-08-03
Proto Updates (Delete key pair + get hub token)

dmerrill6:
# Change log

- Add delete key pair gRPC

- Add delete account gRPC

- Add hub token to GetPublicKey response


dmerrill6:
> LGTM. The other interface is what's needed on the `space-services` side but since that is `sls` maybe it could be discussed on slack?

Just opened a PR for those here https://github.com/FleekHQ/space-services/pull/11/files?short_path=fa8bb37#diff-fa8bb37d54f4833d0ddcc79cb30dbe93

PR #107, opened 2020-08-04
Paper wallet create / restore

dmerrill6:
# Change log

- Implement method to create a keys backup

- Implement method to restore a keys backup

- Add a quick encryption around backup files

- Test flows


core/keychain/test/keychain_test.go
```diff
@@ -0,0 +1,43 @@
...
+	libp2pPriv, _, _ := kc.GetStoredKeyPairInLibP2PFormat()
+
+	// Reset mock store for assertions
+	kc = initTestKeychain(t)
+	mockStore.AssertNotCalled(t, "Set", []byte(keychain.PublicKeyStoreKey), pub)
+	mockStore.On("Set", mock.Anything, mock.Anything).Return(nil)
+
+	kc.ImportExistingKeyPair(libp2pPriv)
+
+	mockStore.AssertCalled(t, "Set", []byte(keychain.PublicKeyStoreKey), pub)
```
perfectmak, line 42: maybe also assert Private Key is set

dmerrill6: Added in b1328bf

PR #108, opened 2020-08-07
Feature/ch17271/hub inbox

jsonsivar:
### Change Log
* Add user client
* Switch thread context between hub vs local
* Setup mailbox on start
* Send message wrapper
* Unit tests


core/textile/client.go
```diff
@@ -18,19 +20,23 @@ import (
...
 	store            db.Store
 	threads          *threadsClient.Client
+	ht               *threadsClient.Client
 	bucketsClient    *bucketsClient.Client
 	isRunning        bool
 	Ready            chan bool
 	cfg              config.Config
 	isConnectedToHub bool
 	netc             *nc.Client
+	uc               UsersClient
```
maurycy, line 39: @jsonsivar 

Do you think it makes to make sense to make the names a bit more descriptive?

core/textile/client.go
```diff
@@ -142,21 +156,78 @@ func (tc *textileClient) start(ctx context.Context, cfg config.Config) error {
...
 	return nil
 }
 
+// adding for testability
+func (tc *textileClient) SetUc(uc UsersClient) {
+	tc.uc = uc
+}
+
+func getUserClient() UsersClient {
+	hubTarget := os.Getenv("TXL_HUB_HOST")
```
maurycy, line 190: @jsonsivar 

Please ignore me if I'm wrong: `TXL_HUB_HOST` or `TXL_HUB_TARGET`? It seems that we remove `TXL_HUB_HOST` from the `set-envs`.

jsonsivar: Good catch. Actually they should come from the global configuration object which is something I forgot to do.

core/textile/client.go
jsonsivar: I'm not too fussed about this but we can change it if we want. Thought `uc` made sense because it stands for user client and in the spirit of Go to keep things short :) (kinda like `netc`/`nc` right above it).

core/textile/client.go
jsonsivar: Fixed in [4d291fd](https://github.com/FleekHQ/space-daemon/pull/108/commits/4d291fdf871f31fde97bd7be5f190b43e72934a1).

core/textile/mailbox.go
```diff
@@ -0,0 +1,45 @@
...
+func (tc *textileClient) SendMessage(ctx context.Context, recipient string, body []byte) (*client.Message, error) {
+	kc := keychain.New(tc.store)
+	var privateKey crypto.PrivKey
+	var err error
+	if privateKey, _, err = kc.GetStoredKeyPairInLibP2PFormat(); err != nil {
+		return nil, err
+	}
+	id := thread.NewLibp2pIdentity(privateKey)
+
+	b := []byte(recipient)
```
dmerrill6, line 27: This is prone to encoding issues. Maybe instead of doing the conversion here, `SendMessage` could receive a `crypto.PubKey` instead of a string.

core/textile/client.go
```diff
@@ -232,16 +304,17 @@ func (tc *textileClient) IsRunning() bool {
...
 
 	if err = tc.requiresRunning(); err != nil {
 		return nil, err
 	}
 
-	// If we are connected to the Hub, add the keys to the context so we can replicate
-	if tc.isConnectedToHub == true {
+	// Some threads will be on the hub and some will be local, this flag lets you specify
+	// where it is, perhaps an improvement could be to save this flag in the store and load it from there
+	if hub {
```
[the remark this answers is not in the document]
jsonsivar: For 1 specifically I thought the metathread would be local but replicated without exposing the data to the hub (so it could be restored later), this way it will still work fully offline. If a hub connected is present when creating the metathread, but then they later go offline it wouldn't work because it wouldn't be able to find even local buckets.  Everything else makes sense to me so will leave it as a flag for the function and remove the comment.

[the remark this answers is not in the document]
dmerrill6: Gotcha, makes sense 👍

core/textile/client.go
```diff
@@ -142,21 +156,79 @@ func (tc *textileClient) start(ctx context.Context, cfg config.Config) error {
...
 	// Attempt to connect to the Hub
-	_, err := tc.getHubCtx(ctx)
+	hubctx, err := tc.getHubCtx(ctx)
 	if err != nil {
 		log.Error("Could not connect to Textile Hub. Starting in offline mode.", err)
 	} else {
 		tc.isConnectedToHub = true
+
+		// setup mailbox
+		mid, err := tc.uc.SetupMailbox(hubctx)
```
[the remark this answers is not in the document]
jsonsivar: Fixed in [e5c0fc5](https://github.com/FleekHQ/space-daemon/pull/108/commits/e5c0fc5a4cf56c80be4552aa84143ad8b4709fc0).

core/textile/client.go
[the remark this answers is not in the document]
jsonsivar: Just cleaned up the comment for now. Didn't add that field to the bucket schema in the metathread yet because all created buckets are local right now (shared buckets aren't added to metathread since it will come from a new shared with me thread).

core/textile/mailbox.go
jsonsivar: Fixed in [e5c0fc5](https://github.com/FleekHQ/space-daemon/pull/108/commits/e5c0fc5a4cf56c80be4552aa84143ad8b4709fc0).

maurycy (approved):

PR #109, opened 2020-08-11
Update GRPC Interfaces

jsonsivar:
Updates include:
* Backup and tentative usage related interfaces
* Updated file info and file info stream


grpc/proto/space.proto
```diff
@@ -245,12 +252,46 @@ service SpaceApi {
...
   }
+
+  rpc ToggleBucketBackup(ToggleBucketBackupRequest) returns (ToggleBucketBackupResponse) {
+    option (google.api.http) = {
+      post: "/v1/backup"
+      body: "*"
+    };
+  }
+
+  rpc GetUsageInfo(GetUsageInfoRequest) returns (GetUsageInfoResponse) {
```
jsonsivar, line 263: This one was had at pending discussion but I added it anyway based on the UI since the implementation detail should be abstracted away by this interface right?

grpc/proto/space.proto
```diff
@@ -260,6 +301,9 @@ message ListDirectoryEntry {
   string updated = 6;
   string fileExtension = 7;
   string ipfsHash = 8;
+  bool local = 9;
```
dmerrill6, line 304: Maybe I'd call this `isLocallyAvailable` or something like that to be more semantic.

grpc/proto/space.proto
```diff
@@ -294,6 +338,7 @@ message Bucket {
   int64 updatedAt = 5;
   repeated BucketMember members = 6;
   bool isPersonalBucket = 7;
+  bool backedUp = 8;
```
dmerrill6, line 341: Maybe `isBackupEnabled`? 🤔

jsonsivar: Fixed in [35ad3a0](https://github.com/FleekHQ/space-daemon/pull/109/commits/35ad3a05949cc481001b673672990921b33a4951).

grpc/proto/space.proto
jsonsivar: Fixed in [35ad3a0](https://github.com/FleekHQ/space-daemon/pull/109/commits/35ad3a05949cc481001b673672990921b33a4951).

PR #111, opened 2020-08-12
Mnemonic backup and restore

dmerrill6:
# Change log

- Add generation and restoring of key pairs by using a bip39 mnemonic.

- Use a deterministic generation of a thread ID for the metathread so that upon restoring the key pair in another machine, we will always end up with the same thread ID for the metathread.

- Clean up and tests.

[ch17742]


core/textile/utils.go
```diff
@@ -23,3 +28,43 @@ func parseDbIDFromString(dbID string) (*thread.ID, error) {
...
 	return &id, nil
 }
+
+type deterministicThreadVariant string
+
+var (
+	metathreadThreadVariant deterministicThreadVariant = "metathread"
+)
+
+func newDeterministicThreadID(st *store.Store, threadVariant deterministicThreadVariant) (thread.ID, error) {
```
perfectmak, line 38: since store is already an interface, you probably don't need to pass it as a pointer again (because then it would be a double pointer to the actual store implementation)

```suggestion
func newDeterministicThreadID(st store.Store, threadVariant deterministicThreadVariant) (thread.ID, error) {
```

core/textile/utils.go
```diff
@@ -23,3 +28,43 @@ func parseDbIDFromString(dbID string) (*thread.ID, error) {
...
+	size := 32
+	variant := thread.Raw
+
+	kc := keychain.New(*st)
```
perfectmak, line 42:
```suggestion
	kc := keychain.New(st)
```

core/textile/utils.go
```diff
@@ -23,3 +28,43 @@ func parseDbIDFromString(dbID string) (*thread.ID, error) {
...
+	// two 8 bytes (max) numbers plus num
+	buf := make([]byte, 2*binary.MaxVarintLen64+numlen)
+	n := binary.PutUvarint(buf, thread.V1)
+	n += binary.PutUvarint(buf[n:], uint64(variant))
+	cn := copy(buf[n:], num)
+	if cn != numlen {
+		panic("copy length is inconsistent")
+	}
+
+	return thread.ID(buf[:n+numlen]), nil
```
perfectmak, line 69: Maybe we need more tests (or comments 😅 ) for this, just to be sure this is sufficient to generate a unique deterministic meta thread id per user.

It look to me like the `buf` buffer is always filled with the same set of value `thread.V1 + variant + numlen` and those values are always static and the same

core/space/services/services_keypair.go
```diff
@@ -5,38 +5,28 @@ import (
...
+	return mnemonic, nil
 }
 
-func (s *Space) generateKeyPairWithForce(ctx context.Context) (domain.KeyPair, error) {
-	kc := keychain.New(s.store)
-	if pub, priv, err := kc.GenerateKeyPairWithForce(); err != nil {
-		return domain.KeyPair{}, err
-	} else {
+func (s *Space) RestoreKeyPairFromMnemonic(ctx context.Context, mnemonic string) error {
+	_, err := s.keychain.GenerateKeyFromMnemonic(keychain.WithMnemonic(mnemonic), keychain.WithOverride())
```
jsonsivar, line 27: Since this can override something they already have setting maybe before restoring, the FE should give a warning that restoring will override the key if already generated right?

core/space/services/services_vault.go
```diff
@@ -65,10 +54,5 @@ func (s *Space) RecoverKeysByLocalBackup(ctx context.Context, path string) error
 		return err
 	}
 
-	// Restore Textile Client state
```
jsonsivar, line 68: Should we still overwrite the thread ID in the store here? Thinking otherwise `findOrCreateMetaThreadID` will return the existing thread ID from an old key if a recovery was done on top of it.

core/textile/utils.go
jsonsivar: I think it's the `num` part that's unique across ID generations.  I think it's expected that the thread ID version and variant are the same while `num` which comes from the privkey is for the uniqueness.

perfectmak: Oh, I just noticed it was `copy(..., num)` and not `copy(..., numlen)`. For some reason, I kept assuming numlen there when I was reviewing

core/keychain/test/keychain_test.go
```diff
@@ -42,3 +46,71 @@ func TestKeychain_GenerateAndRestore(t *testing.T) {
...
+
+	val, err := kc.GenerateKeyFromMnemonic(keychain.WithMnemonic(mnemonic))
+	assert.Nil(t, err)
+	assert.NotNil(t, val)
+	assert.Equal(t, mnemonic, val)
+	mockStore.AssertCalled(t, "Set", []byte(keychain.PublicKeyStoreKey), pubFromMnemonic)
+	mockStore.AssertCalled(t, "Set", []byte(keychain.PrivateKeyStoreKey), privFromMnemonic)
+}
+
+func TestKeychain_RestoreMnemonicKeyOnOverrideErr(t *testing.T) {
```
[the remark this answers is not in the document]
dmerrill6: Added in 4944baf

core/space/services/services_keypair.go
```diff
@@ -5,38 +5,28 @@ import (
 	"encoding/hex"
 
 	"github.com/FleekHQ/space-daemon/core/keychain"
-	"github.com/FleekHQ/space-daemon/core/space/domain"
 	"github.com/FleekHQ/space-daemon/core/textile/hub"
 )
 
-func (s *Space) GenerateKeyPair(ctx context.Context, useForce bool) (domain.KeyPair, error) {
+func (s *Space) GenerateKeyPair(ctx context.Context, useForce bool) (string, error) {
```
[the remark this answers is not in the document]
dmerrill6: Added a comment in 4944baf
I think the new interface is better, since the consumer can choose what to do with the mnemonic, and the key pair didn't offer any use currently, since we might want to avoid exposing the private key and the public key can be fetched with `GetPublicKey`

core/textile/utils.go
dmerrill6: Fixed in 4944baf

core/textile/utils.go
dmerrill6: Fixed in 4944baf

core/textile/utils.go
dmerrill6: Added a test and comments in 4944baf :D

core/keychain/mnemonic.go
```diff
@@ -0,0 +1,86 @@
...
+			return "", err
+		}
+	}
+
+	seed, err := bip39.NewSeedWithErrorChecking(mnemonic, o.password)
+	if err != nil {
+		return "", err
+	}
+
+	compressedSeed := seed[:32]
```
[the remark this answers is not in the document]
dmerrill6: Added a comment in 4944baf
Did this because ed25519 expects the seed to be of size = 32bytes, but the bip39 implementation is fixed to return size=64.
Refactored this so that instead of taking a slice of the seed, we rehash it again but with the desired size.

core/space/services/services_keypair.go
dmerrill6: I think it makes sense to show a warning in the UI, but if you are restoring a key you probably don't care about any existing one, so that's why I defaulted to override. Probably going to change a little bit once we implement allowing multiple keys (https://app.clubhouse.io/terminalsystems/story/17248/app-authorization-flow).

From the current FE perspective, if you are showing the mnemonic input that means the daemon is already running without a key pair, otherwise it would show you as logged in.

core/space/services/services_vault.go
dmerrill6: Changed the imp of `findOrCreateMetaThreadID` in 4944baf to use the key pair as the store key, that way it will keep all threads id and return the correct one for the current key being used.

PR #112, opened 2020-08-14
Embed IPFS node

### LINKED ISSUE #110: v0.0.12 Linux x86_64 build does not have the -ipfsaddr flag
The [readme mentions an `-ipfsaddr` flag](https://github.com/FleekHQ/space-daemon#ipfs-node) that lets the daemon use an existing IPFS node. But the Linux binary doesn't support it:

```
$ ./space_Linux_x86_64 -ipfsaddr=/ip4/127.0.0.1/tcp/5001
flag provided but not defined: -ipfsaddr
Usage of ./space_Linux_x86_64:
  -cpuprofile file
    	write cpu profile to file
  -debug
    	run daemon with debug mode for profiling (default true)
  -dev
    	run daemon in dev mode to use .env file
  -memprofile file
    	write memory profile to file
```

maurycy:
[ch17646]

Fixes #110


core/ipfs/node/node.go
```diff
@@ -0,0 +1,65 @@
...
+
+func NewIpsNode(cfg config.Config) *IpfsNode {
+	return &IpfsNode{
+		Ready: make(chan bool),
+		cfg:   cfg,
+	}
+}
+
+func (node *IpfsNode) Start(ctx context.Context) error {
+    log.Printf("Starting the ipfs node")
```
dmerrill6, line 28: You should import `"github.com/FleekHQ/space-daemon/log"` and then use `log.Debug` instead of printing directly.

core/ipfs/node/node.go
```diff
@@ -0,0 +1,65 @@
...
+
+    proc := process.WithParent(process.Background())
+
+    proc.Go(func(p process.Process) {
+		cmd := exec.Command("space-ipfs-node")
```
dmerrill6, line 33: Should we handle the scenario where the user is running their own IPFS node? Not sure if this should be a flag when starting space daemon (something like `--use-custom-local-ipfs`). What do you think?

jsonsivar: This is definitely something to think about.  Should the daemon also detect if IPFS is running by maybe checking the HTTP API port or the `~/.ipfs` directory and make sure we don't overwrite.

cmd/space-ipfs-node/main.go
```diff
@@ -0,0 +1,154 @@
...
+		Routing: libp2p.DHTClientOption,
+		Repo: repo,
+	}
+
+	node, err := core.NewNode(ctx, nodeOptions)
+	if err != nil {
+		panic(err)
+	}
+
+	core, err := coreapi.NewCoreAPI(node)
```
jsonsivar, line 56: If there is no existing node, we should call the logic behind `ipfs init` to setup the node (ran this and it complained about no ipfs init).

core/ipfs/node/node.go
```diff
@@ -0,0 +1,65 @@
+package ipfs
```
jsonsivar, line 1: So I was able to test the fix from https://github.com/ipfs/go-graphsync/issues/82 so thinking we can try making this embedded again.  I did this in a POC branch here `poc/ipfs-import` but I just copied and pasted to make it quick so prefer you make the edits on this branch to keep commit history good (or I can issue a PR to this branch with the changes).

core/ipfs/node/node.go
maurycy: @dmerrill6 

Great catch!

I just copied `log.Printf` without any consideration from [core/sync/sync.go](https://github.com/FleekHQ/space-daemon/blob/f482e013697206a55fe57cd7450a4f620495cb29/core/sync/sync.go#L100), [changed](https://github.com/FleekHQ/space-daemon/pull/112/commits/14937b1f1d98abd2ba55936c5a1080853602445c) to `log.Info`.

core/ipfs/node/node.go
maurycy: @jsonsivar 

Tried running the branch and got:

```
$ go run ./cmd/space-daemon -dev=true
go: downloading github.com/ipfs/go-ipfs v0.6.1-0.20200817102359-90a573354af2
go: downloading github.com/ipfs/go-ipfs-http-client v0.0.6-0.20200731172059-4e8a6c8b981f
go: downloading github.com/libp2p/go-libp2p-swarm v0.2.8
go: downloading github.com/ipfs/go-ipfs-config v0.9.0
go: downloading github.com/ipfs/go-ipfs-cmds v0.3.0
go: downloading github.com/ipfs/go-graphsync v0.1.1
go: downloading github.com/libp2p/go-libp2p-quic-transport v0.7.1
go: downloading github.com/lucas-clemente/quic-go v0.17.3
go: downloading github.com/prometheus/common v0.10.0
go: downloading github.com/prometheus/procfs v0.1.0
go: downloading github.com/tyler-smith/go-bip39 v1.0.2
panic: qtls.ClientSessionState not compatible with tls.ClientSessionState

goroutine 1 [running]:
github.com/lucas-clemente/quic-go/internal/handshake.init.2()
	/Users/maurycy/go/pkg/mod/github.com/lucas-clemente/quic-go@v0.17.3/internal/handshake/unsafe.go:26 +0x205
exit status 2
```

I think it would be great if we could get a separate pull request, so we keep the conversation (and fixes!) there.

core/ipfs/node/node.go
maurycy: @dmerrill6 @jsonsivar 

What do you think about reusing `ipfsAddr` for this purpose?

```
core/ipfs/ipfs.go
50:	ipfsAddr := cfg.GetString(config.Ipfsaddr, "/ip4/127.0.0.1/tcp/5001")
52:	multiaddress, err := ma.NewMultiaddr(ipfsAddr)
```

It seems to already default to `/ip4/127.0.0.1/tcp/5001` so the implicit assumption of using the local node is here.

core/ipfs/node/node.go
jsonsivar: @maurycy could you try a go mod download and rerun? Because when I run it from that branch I get this
image

Either way, I'll setup a new PR targeting this branch and we can continue discussion there.

maurycy: @jsonsivar 

OK, found the root case: 

https://github.com/lucas-clemente/quic-go/issues/2614

Downgrading.

maurycy: @jsonsivar 

Able to reproduce:

```
[16:43:17] i:maurycy:~/src/space-daemon$ go run ./cmd/space-daemon -dev=true
DEBU[0000] Running mode                                  DevMode=true
DEBU[0000] Running daemon with profiler. Visit http://localhost:6060/debug/pprof
INFO[0000] using path ~/.fleek-space for store
badger 2020/08/18 16:43:27 INFO: All 1 tables opened in 3ms
badger 2020/08/18 16:43:27 INFO: Replaying file id: 0 at offset: 387
badger 2020/08/18 16:43:27 INFO: Replay took: 8.612µs
badger 2020/08/18 16:43:27 DEBUG: Value log discard stats empty
DEBU[0000] Starting Component                            name=Store
DEBU[0000] Starting Component                            name=FolderWatcher
DEBU[0000] Starting Async Component                      name=IpfsNode
INFO[0000] Starting the ipfs node
panic: no IPFS repo found in /Users/maurycy/.ipfs.
please run: 'ipfs init'

goroutine 67 [running]:
github.com/FleekHQ/space-daemon/core/ipfs/node.(*IpfsNode).start(0xc0002d6c60)
	/Users/maurycy/src/space-daemon/core/ipfs/node/node.go:83 +0x5cb
github.com/FleekHQ/space-daemon/core/ipfs/node.(*IpfsNode).Start(0xc0002d6c60, 0x64f98c0, 0xc00014d340, 0x0, 0x0)
	/Users/maurycy/src/space-daemon/core/ipfs/node/node.go:41 +0x40
github.com/FleekHQ/space-daemon/app.(*App).Start.func1(0x0, 0x0)
	/Users/maurycy/src/space-daemon/app/app.go:92 +0x3c
github.com/FleekHQ/space-daemon/app.(*App).RunAsync.func1(0x0, 0x0)
	/Users/maurycy/src/space-daemon/app/app.go:194 +0x26
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc000944000, 0xc0004e04a0)
	/Users/maurycy/go/pkg/mod/golang.org/x/sync@v0.0.0-20200317015054-43a5402ce75a/errgroup/errgroup.go:57 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go
	/Users/maurycy/go/pkg/mod/golang.org/x/sync@v0.0.0-20200317015054-43a5402ce75a/errgroup/errgroup.go:54 +0x66
exit status 2
```

🎉 🎉 🎉 

Let's move the conversation to a new pull request.

core/ipfs/node/node.go
jsonsivar: Yeah sounds like that's a good idea. So if it was already listening on that port it would give an error right?

maurycy: @jsonsivar 

Yes, but... we can check if `config.Ipfsaddr` equals to `"/ip4/127.0.0.1/tcp/5001"` and then, only then, start our node. What do you think?

jsonsivar: Ah I see what you mean. Makes sense to me.

jsonsivar: Not empty though, if its set to localhost we still want to start up the node. Since `core/ipfs/ipfs.go` is the client

maurycy: @jsonsivar 

> Not empty though, if its set to localhost we still want to start up the node.

But what if the user wants to start his own node? 

My reasoning: if the user _explicitly_ asks us to connect to the user node (ie: by setting `config.Ipfsaddr` to `/ip4/127.0.0.1/tcp/5001`) then we should respect that.

Otherwise (ie: if `config.Ipfsaddr` is empty), we should start our own node and set `config.Ipfsaddr` accordingly (ie: `/ip4/127.0.0.1/tcp/5001` by default) since it's used by `core/ipfs/ipfs.go` client, and `space` actually becomes both server and client now.

Any node configuration should use different configuration (`config.IpfsServerAddr`) that would eventually set `config.Ipfsaddr` if blank.

jsonsivar: That makes a lot of sense :+1:

jsonsivar:
@dmerrill6 @maurycy since we merged quite a bit of changes to this since the overall approach changed, think it would be a good idea for all of us to review again before going in to master.

PR #115, opened 2020-08-18
OS keychain

dmerrill6:
# Change log

- Store private key in OS provided keychains instead of the store

- Refactor keychain access across the app to reduce the amount of places where we fetch from the keychain (public key is stored in store which is easier to access, so updated all places that were getting the private key and deriving the public key to get the public key directly instead).

- Mocked the keyring lib for testing the new keychain

- Fixed broken tests

[ch17842]


core/keychain/keychain.go
```diff
@@ -2,40 +2,98 @@ package keychain
...
+
+type Option func(o *keychainOptions)
+
+func New(opts ...Option) *keychain {
+	o := defaultKeychainOptions
+	for _, opt := range opts {
+		opt(&o)
+	}
+
+	defaultStore := store.New(store.WithPath(o.fileDir))
```
jsonsivar, line 87: Should we move this into the `if` below so we don't need to create it if it was already supplied?

core/keychain/keychain.go
```diff
@@ -95,7 +135,37 @@ func (kc *keychain) GetStoredKeyPairInLibP2PFormat() (crypto.PrivKey, crypto.Pub
...
 // Warning: If there's already a key pair stored, it overrides it.
 func (kc *keychain) GenerateKeyPairWithForce() ([]byte, []byte, error) {
-	return kc.generateAndStoreKeyPair(nil)
+	return kc.generateAndStoreKeyPair(nil, "")
+}
+
+// Returns the public key currently in use in LibP2P format.
+// Returns an error if there's no public key set.
+// Unlike GetStoredKeyPairInLibP2PFormat, this method does not access the keychain
+func (kc *keychain) GetStoredPublicKey() (crypto.PubKey, error) {
```
jsonsivar, line 144: Should this be checking the keyring directly in addition to the store?

core/keychain/keychain.go
```diff
@@ -137,7 +203,7 @@ func (kc *keychain) generateKeyPair(seed []byte) ([]byte, []byte, error) {
 	return pub, priv, err
 }
 
-func (kc *keychain) generateAndStoreKeyPair(seed []byte) ([]byte, []byte, error) {
+func (kc *keychain) generateAndStoreKeyPair(seed []byte, mnemonic string) ([]byte, []byte, error) {
```
jsonsivar, line 206: Can someone pass in an uncorrelated `seed` and `mnemonic` here? Maybe it just gets one and the other is generated so there is less room for error? Edit: it could be updated later too since we only use it in one place rn.

core/keychain/keychain.go
dmerrill6: Fixed in 8d0b7d5

core/keychain/keychain.go
dmerrill6: Did that on purpose, that way if you have strict settings for your keychain, we don't have to be prompting for system's password whenever we try to access this public info (this was one of the optimizations so that textile client doesn't access the keychain every time it accesses a thread).

core/keychain/keychain.go
dmerrill6: This is an auxiliary/private method called after generating the mnemonic / seed so feel it's fair to assume it's not going to be misused. Can't generate one from the other since it's being used when generating keys without mnemonic. Agree we need to update it later since generating keys without mnemonic might be deprecated and we can clean up the keychain class.

dmerrill6:
> Resolved comments. Updates LGTM and thanks for updating the mailbox tests.

Thanks for the quick review! Merging...

PR #118, opened 2020-08-20
GRPC Updates: create share API, and add get shared with me files

jsonsivar:


grpc/proto/space.proto
```diff
@@ -199,9 +206,9 @@ service SpaceApi {
   }
 
   // Share bucket via public key using Textile Hub inboxing
-  rpc ShareBucketViaPublicKey(ShareBucketViaPublicKeyRequest) returns (ShareBucketViaPublicKeyResponse) {
+  rpc ShareFilesViaPublicKey(ShareFilesViaPublicKeyRequest) returns (ShareFilesViaPublicKeyResponse) {
```
jsonsivar, line 209: Changed this instead of adding a new one since we wont be sharing buckets as a whole anytime soon (unless I'm missing something).

giankotarola:
lgtm

grpc/proto/space.proto
```diff
@@ -232,15 +239,6 @@ service SpaceApi {
     };
   }
 
-  // Copies a set of files in a new or existing bucket that's shared with a group of public keys
-  // Auxiliary method that internally calls CreateBucket, ShareBucketViaPublicKey, AddFile.
-  rpc CopyAndShareFiles(CopyAndShareFilesRequest) returns (CopyAndShareFilesResponse) {
```
jsonsivar, line 237: Again we won't need this given how the sharing architecture worked out to be but let me know if its better we leave it for future uses (can't think of any).

maurycy (approved):
LGTM. 🚀

PR #119, opened 2020-08-20
Hotfix: Error on startup due to hub context issue

jsonsivar:
Refactored temporary hub context helper to return context instead of token since it requires the API sig context embedded into it as well.


maurycy (approved):

maurycy:
This is the error mentioned on Slack yesterday?

jsonsivar:
> This is the error mentioned on Slack yesterday?

No that was due to missing env vars. This was just a regression from a little hub context refactoring.

PR #121, opened 2020-08-24
Vault: Clientside implementation

dmerrill6:
# Change log

- Add vault docs that explains the security protocol of the vault

- Add vault class, which implements the protocol

- Add tests for the vault class

## TODO:

Connect to gRPC method and test integration

Will remove draft state of this PR once this is tested end to end.


core/vault/vault.go
```diff
@@ -0,0 +1,254 @@
...
+	storeRequest := &storeVaultRequest{
+		Vault: base64.RawStdEncoding.EncodeToString(encVf),
+		Vsk:   base64.RawStdEncoding.EncodeToString(vsk),
+	}
+	reqJSON, err := json.Marshal(storeRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := http.Post(
```
studna, line 100: should we use `appToken` as authorization header, so only auth users are allowed to post keys?

core/vault/vault.go
```diff
@@ -0,0 +1,254 @@
...
+	// Send retrieve request to vault service
+	reqJSON, err := json.Marshal(&retrieveVaultRequest{
+		Vsk: base64.RawStdEncoding.EncodeToString(vsk),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := http.Post(
+		v.vaultAPIURL+"/vaults/"+uuid,
```
studna, line 134: I assume uuid is obtained by `getIdentityByUsername`, right? Maybe it'd be better if we can use username here directly, so we can secure `getIdentityByUsername` endpoint with app JWT (rate limits on token) and this one differently (like IP-rate limited).

studna:
Another idea - maybe it'd be better if we return some randomly generated response even if `vsk` does not match our records. So in case attacker guess `vsk`, he won't be able to if `vsk` is valid without knowing the password.

On the other hand cracking `vsk` would take some time, so maybe it's just overengineering.

core/vault/vault.go
dmerrill6: Yea, adding that when testing the integrated flow.

core/vault/vault.go
dmerrill6: Don't want to use username for the following reasons:
- Vault endpoint shouldn't change if the user changes their username
- Daemon doesn't deal with usernames. It shouldn't deal with uuids, but at least these are immutable registries that map to a vault. So if we migrate the identity provider to something like Ceramic later on, we can keep using these uuids as an attached document.

I don't see any strong reasons for securing the `getIdentityByUsername` endpoint. These are public registries anyways 🤔

dmerrill6:
> Another idea - maybe it'd be better if we return some randomly generated response even if `vsk` does not match our records. So in case attacker guess `vsk`, he won't be able to if `vsk` is valid without knowing the password.
> 
> On the other hand cracking `vsk` would take some time, so maybe it's just overengineering.

One of the reasons we do this cryptographic flow is that this way we allow users to know if the password they are inputting is right or wrong. If we always return a random `vk(vf)` wouldn't the user be logged into a new account instead of knowing they have to retype their password?

core/vault/vault.go
studna: Oh I see, ok. Initially I thought securing that endpoint would make harder to scan users, like you can have a list of twitter accounts (github, etc.) and you can scan our endpoint and reveal that user `amanda93` is using space storage. Then you can use this information for phishing attacks and so on (aka "Your documents leaked from your space storage!")..

dmerrill6: Gotcha, I see what you are saying. On the other hand, it's very easy for this scammer to create an account first and then use their JWT before calling the address book, so this would be a very small barrier.

studna:
> One of the reasons we do this cryptographic flow is that this way we allow users to know if the password they are inputting is right or wrong. If we always return a random vk(vf) wouldn't the user be logged into a new account instead of knowing they have to retype their password?

He would know he's not able to decrypt the file with used password. I mean it only hardens brute forcing about one more step. But since we'll have API Rate limiting it might be unnecessary complexity - I think even without this step, brute forcing vsk -> password would take so many years.

dmerrill6:
> > One of the reasons we do this cryptographic flow is that this way we allow users to know if the password they are inputting is right or wrong. If we always return a random vk(vf) wouldn't the user be logged into a new account instead of knowing they have to retype their password?
> 
> He would know he's not able to decrypt the file with used password. I mean it only hardens brute forcing about one more step. But since we'll have API Rate limiting it might be unnecessary complexity - I think even without this step, brute forcing vsk -> password would take so many years.

Exactly. Also the server side will be running 1 million KDF iterations so it takes about 1 second for each attempt 😂

core/vault/vault.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	gcm, err := cipher.NewGCM(block)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(ciphertext) < gcm.NonceSize() {
```
maurycy, line 222: @dmerrill6 

Only `<` or `!=` ?

core/vault/vault.go
```diff
@@ -0,0 +1,254 @@
...
+	)
+}
+
+func parseAPIResponse(resp *http.Response) ([]byte, error) {
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
```
maurycy, line 239: @dmerrill6 

Or just `!= 200` (and, possibly, few others)?

maurycy (approved):
LGTM. 🚀

core/vault/vault.go
dmerrill6: Added in 8587445

dmerrill6:
Implemented the service layer that stores and backups the current key pair + mnemonic. Marking this PR as no longer a draft.

core/vault/vault.go
dmerrill6: I think `<` is correct since, if I understand correctly, the nonce is prepended to the beginning of the ciphertext therefore it's fine if `len(ciphertext) >= nonce size`

core/vault/vault.go
dmerrill6: Just wanted to leave this flexible and not hardcode all possible success codes (200, 201 = created, etc.)

PR #122, opened 2020-08-25
Partial release of share files implementation

jsonsivar:
Figured we could release this partial implementation of the share files via pubkeys so that we can test the integrations and notifications.  Once textile PR is merged we can update placeholder with call to buckets ACL api.

Re: tests, send notification was unit tested in another PR and think we can add tests for the grpc/service once we finalize the interface for buckets ACL api.


core/space/services/services_sharing.go
```diff
@@ -222,3 +224,39 @@ func (s *Space) OpenSharedFile(ctx context.Context, hash, password, filename str
 		Location: decryptedFile.Name(),
 	}, nil
 }
+
+func (s *Space) ShareFilesViaPublicKey(ctx context.Context, bucketName string, paths []string, pubkeys []crypto.PubKey) error {
```
perfectmak, line 228: Just out of curiosity, how fast is this function overall? Especially, if we have quite a number `pubkeys` we want to share the files with.

perfectmak (approved):
LGTM

core/space/services/services_sharing.go
```diff
@@ -222,3 +224,39 @@ func (s *Space) OpenSharedFile(ctx context.Context, hash, password, filename str
...
+	err := s.tc.ShareFilesViaPublicKey(ctx, bucketName, paths, pubkeys)
+	if err != nil {
+		return err
+	}
+	for _, pk := range pubkeys {
+		err := s.tc.SetMember(ctx, bucketName, pk, nil)
```
jsonsivar, line 234: Actually, this has to change because membership is now at the file level so doesn't make sense to add at bucket level

core/space/services/services_sharing.go
jsonsivar: That's a good point, I guess its going to be as slow as much as there are pubkeys.  Small update on this though I think we can update the `SetMembers` interface to take the collection of pubkeys so we can just have one call there.  And re: the messaging that is temporary since it will be within the bucket ACL api textile is adding but think we can wait to see performance once the actual implementation is done.

core/textile/collections.go
```diff
@@ -230,3 +235,95 @@ func (tc *textileClient) initBucketCollection(ctx context.Context) (context.Cont
...
+
+	pk, err := crypto.MarshalPublicKey(p2ppk)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	return pk, nil
+}
+
+func (tc *textileClient) SetMember(ctx context.Context, slug string, pk crypto.PubKey, msg *string) error {
```
maurycy, line 279: @jsonsivar 

What do you think about taking a new `struct` as a param?

core/textile/collections.go
```diff
@@ -230,3 +235,95 @@ func (tc *textileClient) initBucketCollection(ctx context.Context) (context.Cont
...
+	}
+	bctx, dbID, err := tc.GetBucketContext(ctx, slug)
+	_, err = tc.threads.Create(bctx, *dbID, membersCollectionName, client.Instances{m})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (tc *textileClient) SetOwner(ctx context.Context, slug string) error {
```
maurycy, line 312: @jsonsivar 

Do we need `GetOwner`?

core/textile/collections.go
```diff
@@ -230,3 +235,95 @@ func (tc *textileClient) initBucketCollection(ctx context.Context) (context.Cont
...
+	pk, err := tc.getSessionKeyAsNativePubKey()
+	if err != nil {
+		return err
+	}
+
+	m := &domain.Member{
```
maurycy, line 318: @jsonsivar 

Do I correctly understand that the difference between owner and member is that the former has no `InvitationID` etc.?

maurycy (approved):

jsonsivar:
> Overall looks good but not fully understanding the collection updates, as those look like they are designed to set and track the members of a bucket, but not at the file level.

Yeah I added a comment highlighting that because it was bucket level before but since we changed it to file level ACL we need to update the get/set members.

> Some thoughts:
> 
>     * Given that the collections.go file is growing too much, maybe we can create a `models` folder and have a file per collection.

That works.

>     * I'd change the method names to be a bit more redundant. Given the `TextileClient` is growing to have a lot of methods, if we are very clear with the names it'd be easier to use. An example would be `SetFileMember` instead of just `SetMember`.

That makes sense.

>     * Not sure where ACL fits in, but wondering what happens if a non owner calls `SetOwner` or stuff like that. Are there unhandled edge-cases where collections get out of sync with the real ACL? Should we handle this?

So this is strictly metadata that is to be used for our screens to show pending invites/acceptances and we needed to distinguish between owner and members for our own UI. The actual ACL and access control logic is going to be built into threads/buckets within textile libs.  I was debating if we still need this given that, but since this is a placeholder implementation until we integrate to textile left it so we can adjust based on what we can/cannot do with the threads ACL list (e.g., that schema won't have whether someone has accepted a share invite or not). We can probably even get rid of the owner concept at this level because we only needed it to separate between My Files and Shared With Me files when we had the old bucket-per-share design. 

>     * Related to the above point, maybe an option is keeping `SetOwner`, `SetMember`, etc. private. So the Space service calls `ShareFilesViaPublicKey`, and internally `TextileClient` updates what it needs to update. That way only `GetMembers` is public. Feels like a better isolation of concern as the consumer of the TextileClient doesn't need to worry about internals and syncing.

Yeah these could be private functions but we might also be able to get rid of some of them if we don't need to save owner on the this collection anymore.  Might change this model to be `Invites` instead of of `Members` so it's not confusing that it might entail access control logic.

maurycy:
@dmerrill6 

I agree with the `models`. My intuition is that we started doing a CRUD / MC (without the V) implicitly over the past few pull requests.

PR #123, opened 2020-08-25
Feature:  Bucket Level Encryption

perfectmak:
**Update**
Now instead of path level encryption keys, we maintain a single encryption key for a bucket and that is re-used to encrypt all paths and their file contents before pushing to IPFS.

**CHANGELOG**
- Add EncryptionKey to bucket schema
- Abstraction around encrypting and decrypting bucket items
- Encrypt and Decrypt ListPath data [ch17961]
- Encrypt and Decrypt PushPath data [ch17942] & [ch17955]
- Encrypt and Decrypt RemovePath data [ch17968]
- Decrypt PullPath data before writing it back [ch17966]

Test include a unit test for the core encryption and decryption logic and for now did manual tests for the critical paths:
- Create a new bucket
- Create a directory
- List created directories and verify contents
- Uploaded a local file
- Fetched/Pulled the file from remote and it shows decrypted locally
- Also fetched the IPFS content hash for files and see that they are encrypted on IPFS

### DIFF
--- a/core/textile/bucket/crypto/crypto.go
+++ b/core/textile/bucket/crypto/crypto.go
@@ -1,0 +1,174 @@
+package crypto
+
+import (
+	"bytes"
+	b64 "encoding/base64"
+	"errors"
+	"io"
+	"io/ioutil"
+	"strings"
+
+	"github.com/FleekHQ/space-daemon/log"
+)
+
+func parseKeys(key []byte) (aesKey, iv, hmacKey []byte, err error) {
+	if len(key) != aesKeySize+ivKeySize+hmacKeySize {
+		return nil, nil, nil, errors.New("unsupported encryption keys provided.")
+	}
+
+	return key[:aesKeySize], key[aesKeySize:(aesKeySize + ivKeySize)], key[(aesKeySize + ivKeySize):], nil
+}
+
+// EncryptPathItems returns an encrypted path and a Reader that reads the encrypted data from the
+// plain reader passed into the function.
+// Encrypted data is AES-CTR of data + AES-512 HMAC of encrypted data
+//
+// NOTE: key must be a 64 byte long key
+// To decrypt the result of this function use the DecryptPathItems function
+func EncryptPathItems(key []byte, path string, plainReader io.Reader) (string, io.Reader, error) {
+	// split key into key and secret
+	// use key and secret and IV
+
+	// encrypt path
+	aesKey, iv, hmacKey, err := parseKeys(key)
+	if err != nil {
+		return "", nil, err
+	}
+
+	encryptedPath := ""
+	pathParts := strings.Split(path, "/")
+	pathPartsLen := len(pathParts)
+	for i, pathItem := range pathParts {
+		if pathItem == "" {
+			continue
+		}
+		encryptedPathReader, err := NewEncryptReader(
+			strings.NewReader(pathItem),
+			aesKey,
+			iv,
+			hmacKey,
+		)
+		if err != nil {
+			return "", nil, err
+		}
+
+		encryptedPathItem, err := readAsBase64Strings(encryptedPathReader)
+		if err != nil {
+			return "", nil, err
+		}
+		encryptedPath += encryptedPathItem
+		if i != pathPartsLen-1 {
+			encryptedPath += "/"
+		}
+	}
+
+	// encrypt data
+	var encryptedReader io.Reader
+	if plainReader != nil {
+		var err error
+		encryptedReader, err = NewEncryptReader(
+			plainReader,
+			aesKey,
+			iv,
+			hmacKey,
+		)
+		if err != nil {
+			return "", nil, err
+		}
+	}
+
+	return encryptedPath, encryptedReader, nil
+}
+
+// DecryptPathItems returns a decrypted path string and an io.Reader that reads the decrypted data from the
+// encrypted reader passed into the function.
+// To only decrypt a path, pass in an empty byte buffer as the reader and check only for the string result.
+//
+// NOTE: key must be a 64 byte long key
+func DecryptPathItems(key []byte, path string, encryptedReader io.Reader) (string, io.ReadCloser, error) {
+	// decrypt path
+	log.Debug("Decrypting Path Items", "path:"+path)
+	aesKey, iv, hmacKey, err := parseKeys(key)
+	if err != nil {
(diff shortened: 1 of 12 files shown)


core/textile/collections.go
```diff
@@ -230,3 +248,84 @@ func (tc *textileClient) initBucketCollection(ctx context.Context) (context.Cont
...
+func (tc *textileClient) upsertBucketPathEncryption(
+	ctx context.Context,
+	bucketSlug, path, key string,
+) (*BucketPathEncryptionSchema, error) {
+	metaCtx, metaDbID, err := tc.initBucketCollection(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	existingPathEncryption, err := tc.findBucketPathEncryption(ctx, bucketSlug, path)
```
maurycy, line 310: @perfectmak 

What do you think about `cleanBucketPath(path)` here?

maurycy (approved):

core/textile/collections.go
```diff
@@ -230,3 +248,84 @@ func (tc *textileClient) initBucketCollection(ctx context.Context) (context.Cont
...
+	res, err := tc.threads.Find(
+		metaCtx,
+		*metaDbID,
+		pathEncryptionCollectionName,
+		db.Where("bucket_slug").Eq(bucketSlug).And("_id").Eq(path),
+		&BucketPathEncryptionSchema{},
+	)
+
+	if err != nil {
+		log.Debug("Failed to find bucket path encryption", "error:"+err.Error(), "bucket:"+bucketSlug, "path:"+path)
```
dmerrill6, line 289: Maybe you can use log.Error

core/textile/collections.go
perfectmak: Makes sense. I'll use it here as well.

core/textile/collections.go
perfectmak: Yeah, I don't know why I used debug here 😅

core/textile/collections.go
perfectmak: Actually, on second thought, I'll be cleaning the path from the public functions that would be using the private `upsertBucketPathEncryption()` function, so no need to double clean it here.

jsonsivar:
>There is still an issue though with the encryption. The textile dcrypto encryption logic doesn't do a deterministic encryption because of the nature of AES-CTR and random seed it uses, so this means that subsequent re-encryption of the same path for future request ends up producing a different value and textile will not be able to find the path in ipfs.

Just to continue the discussion here, thinking these are the options we have then:
1. Request Textile to expose the seed field
2. Use our own encryption library where we can save and retrieve the random seeds

In the meantime going to look at the dcrypto lib and see if there are any other ideas.

core/textile/bucket/crypto/crypto.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// encrypt path
+	log.Debug("Encrypting Path Items", "path:"+path)
+	aesKey, iv, hmacKey, err := parseKeys(key)
+	if err != nil {
+		return "", nil, err
+	}
+
+	encryptedPath := ""
+	pathParts := strings.Split(path, "/")
```
dmerrill6, line 40: Are these paths bucket paths or os paths? Because if it's the latter maybe it should use `path/filepath` package to use cross-platform path delimiters

core/textile/bucket/crypto/crypto.go
```diff
@@ -0,0 +1,177 @@
...
+		if err != nil {
+			return "", nil, err
+		}
+		encryptedPath += encryptedPathItem
+		if i != pathPartsLen-1 {
+			encryptedPath += "/"
+		}
+	}
+
+	log.Debug("Encrypting Path Items", "path:"+path, "encryptedPath:"+encryptedPath)
```
dmerrill6, line 66: Maybe we should remove this log when merging because it defeats the purpose of hiding data through encryption.

core/textile/bucket/crypto/decrypter.go
```diff
@@ -0,0 +1,98 @@
...
+// NewDecryptReader creates an io.ReadCloser wrapping an io.Reader using the keys and iv
+// to decode the content using AES and verify HMAC.
+func NewDecryptReader(r io.Reader, aesKey, iv, hmacKey []byte) (io.ReadCloser, error) {
+	return newDecryptReader(r, aesKey, iv, hmacKey)
+}
+
+func newDecryptReader(r io.Reader, aesKey []byte, iv []byte, hmacKey []byte) (io.ReadCloser, error) {
+	mac := make([]byte, hmacSize)
+	h := hmac.New(hashFunc, hmacKey)
+	dst, err := tmpfile.New(&tmpfile.Context{
```
dmerrill6, line 28: Is there a way we can decrypt to memory instead of to disk?

dmerrill6 (approved):
LGTM, left a few comments/questions though

core/textile/bucket/crypto/decrypter.go
dmerrill6: Maybe for big files this is not ideal though so does this at least store the encrypted files to disk and not the decrypted ones?

core/textile/bucket/crypto/crypto.go
perfectmak: True, I'll remove it

core/textile/bucket/crypto/crypto.go
perfectmak: I'm assuming at this level we are dealing with bucket paths, so all transformation should have been done before a request is passed down to the buckets, since non unix `/` paths don't work with bucket I'm not using the filepath or os specific path separator so it doesn't fail for other operating systems

core/textile/bucket/crypto/decrypter.go
perfectmak: Only encrypted content is written to the file... It is while it is being read that it gets decrypted.

dmerrill6: Is this file deleted after it finished reading?

perfectmak: Yes it is. Once the Done() function is called on the `dst`. That is why that special tmpfile package is used.

dmerrill6: Nice

PR #124, opened 2020-08-25
fix(docs): devMode => dev in setup

andrewxhill:


maurycy (approved):
@andrewxhill Great catch!

PR #125, opened 2020-08-26
Update proto schema: GetRecentlySharedWith

dmerrill6:
# Change log

- Added GetRecentlySharedWith method

- Updated Vault grpc (Backup/Restore keys by passphrase) to avoid conflicts later when merging #121 

Created a card to track the implementation of that method here [ch18024]


dmerrill6:
> LGTM. Just a note on the address: I had originally left the `address` property on the `FileMember` because the plan was to track it in the metathread but going forward we will pull from the bucket ACL itself in the hub which doesn't have the concept of `address`.

That shouldn't be a problem right? We just pass all pub keys through a mapping function to derive address before returning to the client.

PR #127, opened 2020-08-27
Hotfix: use cfg values for txl hub key and secret

jsonsivar:
Was investigating @gpuente 's issue on the binary failing if the envs weren't there which was incorrect since they should already be injected into the binary.  Looks like these 2 fields were still being brought in by env vars directly instead of the config file.


maurycy (approved):
LGTM. 🚀

PR #128, opened 2020-08-27
Textile Hub auth challenge

studna:


core/textile/hub/hub_auth.go
```diff
@@ -63,26 +68,27 @@ func storeHubToken(st store.Store, hubToken string) error {
...
-	if valFromStore, err := getHubTokenFromStore(st); err != nil {
-		return "", err
-	} else if valFromStore != "" {
-		log.Debug("Got hub token from store: " + valFromStore)
-		return valFromStore, nil
-	}
+	// if valFromStore, err := getHubTokenFromStore(st); err != nil {
+	// 	return nil, err
+	// } else if valFromStore != "" {
+	// 	log.Debug("Got hub token from store: " + valFromStore)
```
dmerrill6, line 76: We should reenable this, and add an expiry check before returning so that we can refresh.

core/textile/hub/hub_auth.go
```diff
@@ -140,11 +149,26 @@ func GetHubToken(ctx context.Context, st store.Store, kc keychain.Keychain, cfg
...
-	return token.Value.Token, nil
+	ctx = common.NewAPIKeyContext(ctx, token.Value.Key)
+	ctx = common.NewAPISigContext(ctx, token.Value.Msg, sig)
+
+	tok := thread.Token(token.Value.Token)
+	ctx = thread.NewTokenContext(ctx, tok)
+
+	// @todo: Store appToken
+
+	// this might not be ideal, we need to store also key + msg + sig
```
dmerrill6, line 166: I think it's better to store it to avoid doing this slow challenge too many times

PR #129, opened 2020-08-27
Log out gRPCs

dmerrill6:
# Change log

When logging out, we need to let the user test their password to see if they remember it correctly and then delete the keys. This PR implements those two methods.

[ch18081]


grpc/proto/space.proto
```diff
@@ -181,6 +181,14 @@ service SpaceApi {
     };
   }
 
+  // Tests a passphrase to see if it matches the one previously used
+  rpc TestKeysPassphrase(TestKeysPassphraseRequest) returns (TestKeysPassphraseResponse) {
```
perfectmak, line 185: I have a question about the possibility of the `TestKeysPassphrase` RPC being misused and possibly be used to brute-force and determine a users password? Although, I'm not yet sure what an alternative would be, perhaps we should put some sort of rate-limit on that rpc call (like after 3 trials you wait a day)

perfectmak (approved):
LGTM!

grpc/proto/space.proto
dmerrill6: Check out `docs/crypto/vault` file, it outlines the security measures to prevent that.

To summarize:
- It takes 1 second per attempt
- We will add IP based rate limiting

In the future we will also have "app tokens" so that the daemon port is not open and can only be accessed by authorized apps. That way as long as the vault service is secured we should be fine.

PR #132, opened 2020-08-27
Hotfix/embedded ipfs node: in dev mode the repo path was being set to emtpy

jsonsivar:
The `cfg.GetString` logic checks if the key exists so it doesn't block out empty strs.  In dev mode, it was being set to empty via a missing env vars.


maurycy (approved):

jsonsivar:
Closing in favour of #133

PR #134, opened 2020-08-28
Feature/ch17805/get notifs and listen

jsonsivar:
### Change Log
* Added get notifications and listener helpers
* Parsers and formatters for mail
* Updated to use local mailbox lib so we can watch

### TODO:
- mocked tests


app/app.go
```diff
@@ -172,6 +172,8 @@ func (a *App) Start(ctx context.Context) error {
 		grpc.WithRestProxyPort(a.cfg.GetInt(config.SpaceRestProxyServerPort, 0)),
 	)
 
+	textileClient.ListenForMessages(ctx, srv)
```
dmerrill6, line 175: Maybe this can be started in the healthcheck / start function of textile client

core/textile/client.go
```diff
@@ -172,6 +178,52 @@ func (tc *textileClient) SetUc(uc UsersClient) {
...
+	mbpath := filepath.Join(usr.HomeDir, ".fleek-space/textile/mail")
+
+	var mailbox *mail.Mailbox
+	dbid, err := tc.store.Get([]byte(mailboxSetupFlagStoreKey))
+	if err == nil && len(dbid) > 0 {
+		// restore
+		mailbox, err = maillib.GetLocalMailbox(ctx, mbpath)
+		if err != nil {
+			return err
+		}
```
dmerrill6, line 198: Maybe `return mailbox, nil`  here to avoid the `else`.

Or you can wrap this into a func like `GetMailboxWithCache`.

Also wondering if some of this init logic can be moved to `mailbox.go`.

core/textile/client.go
```diff
@@ -172,6 +178,52 @@ func (tc *textileClient) SetUc(uc UsersClient) {
 	tc.uc = uc
 }
 
+func (tc *textileClient) SetMb(mb Mailbox) {
```
dmerrill6, line 181: I see this is only used for tests, but feels wrong adding more stuff to the textile client public interface. Maybe you can change the package name of `textile_test.go` to `textile` to get access to private fields and simply do `tc.mb = mockMb`. Or maybe it's not too bad as long as `textile.go` doesn't expose that method.

dmerrill6 (approved):
LGTM, left some minor stuff.

app/app.go
jsonsivar: For this I needed to pass in `srv` which is not currently exposed in the textile client (this is so that the message handler can issue a grpc stream event when a message is received). Also we start the grpc server after the textile client so `srv` isn't available when initing textile client :exploding_head:

core/textile/client.go
jsonsivar: Fixed in [5a39569](https://github.com/FleekHQ/space-daemon/pull/134/commits/5a3956938857a5752d199d3ea9223ca4ca983837). Note that the `else` was one level above so I wasn't sure how to remove it.  However I moved the mailbox init into mailbox.go and refactored into 2 smaller functions.

core/textile/client.go
jsonsivar: Fixed in [4f496ce](https://github.com/FleekHQ/space-daemon/pull/134/commits/4f496cebd9add9c7d6cd86c2194408a4995bf3a3). I removed them both by making them init parameters and injecting them from the top.  The helpers to create the clients still in textile package but exposed one of them so we can call it from `app.go` and inject.  Also the weird part is I made `mb` optional in the init because it gets set later, but this way I can inject during test as well. Not perfect but better than before I think.

dmerrill6 (approved):
LGTM

PR #137, opened 2020-08-29
don't block <-component.WaitForReady() on err

maurycy:
The issue is that any `RunAsync()` `func` silently blocks the `app.Start()` on error, ie:

```golang
func (a *App) RunAsync(name string, component core.AsyncComponent, fn func() error) {
[...]

	a.eg.Go(func() error {
		return fn()
	})

	<-component.WaitForReady()
```

but:

```golang
func (tb *TextileBuckd) Start(ctx context.Context) error {
[...]

	usr, err := user.Current()
	if err != nil {
		return err
	}

[...]

	tb.IsRunning = true
	tb.Ready <- true
	return nil
}
```

Also, we're not showing any error returned by `RunAsync()` `func`.

The pull request aims to fix both.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -97,24 +97,34 @@ func (a *App) Start(ctx context.Context) error {
 	if a.cfg.GetBool(config.Ipfsnode, true) {
 		// setup local ipfs node
 		node := node.NewIpsNode(a.cfg)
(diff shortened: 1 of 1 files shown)


perfectmak (approved):
Good catch. Thanks for this, it looks good to me

PR #138, opened 2020-08-31
getMetaThreadContext: pass hub

maurycy:
Not 100% sure if this is correct.


jsonsivar:
> If you check the comments in #108 I think this is set to false on purpose. @jsonsivar can further comment

Yeah agreed. It should be always false since the main meta thread will always be local. Only time it will be remote is during a restore scenario but then the user will join the thread info and it becomes local anyways.

Having said that, I see why the fix was made because the parent function takes in a bool to specify hub or not, so maybe the fix is on that lin.  I.e.: to this

```
func (tc *textileClient) getMetaThreadContext(ctx context.Context) (context.Context, *thread.ID, error) {
```

PR #141, opened 2020-09-01
Allow running Textile client in no-key state

dmerrill6:
# Change log

- Textile client is now started on boot but does not require a key. Initialization is delayed until there's key

- Message channel so that textile can initialize defaults when key is added, and to reset to default state when key is removed.

- Fixed delete key response

[ch17853]


app/app.go
```diff
@@ -97,9 +97,12 @@ func (a *App) Start(ctx context.Context) error {
...
-			return node.Start(ctx)
+		a.RunAsync("IpfsNode", node, func() error {
+			if err := node.Start(ctx); err != nil {
+				log.Error("Error starting embedded IPFS node", err)
+				return err
+			}
+			return nil
 		})
 		if err != nil {
 			return err
```
maurycy, line 108: @dmerrill6 

What do you think about adding `log.Error("Error starting embedded IPFS node", err)` here?

It should be the same but does not break the `err = a.RunAsync()` pattern.

core/textile/client.go
```diff
@@ -125,7 +129,28 @@ func (tc *textileClient) start(ctx context.Context, cfg config.Config) error {
 	tc.ConnectToHub(ctx)
 
 	tc.Ready <- true
-	return nil
+
+	// Healthcheck flow
+	tc.healthcheck(ctx)
+	for {
+		timeAfterNextCheck := 60 * time.Second
+		if tc.isConnectedToHub == false || tc.isInitialized == false {
```
maurycy, line 137: @dmerrill6 

`tc.requiresRunning()`?

core/textile/client.go
```diff
@@ -125,7 +129,28 @@ func (tc *textileClient) start(ctx context.Context, cfg config.Config) error {
...
+			timeAfterNextCheck = 5 * time.Second
```
maurycy, line 138: @dmerrill6 

Do you think there's a risk of flooding the hub with so many so frequent connections?

core/textile/client.go
```diff
@@ -198,35 +236,9 @@ func getHubThreadsClient(host string) *threadsClient.Client {
 	return tc
 }
 
-// StartAndBootstrap starts a Textile Client and also initializes default resources for it like a key pair and default bucket.
-func (tc *textileClient) Start(ctx context.Context, cfg config.Config) error {
-	// Create key pair if not present
-	log.Debug("Starting Textile Client: Getting key status...")
-	_, err := tc.kc.GetStoredPublicKey()
+func (tc *textileClient) initialize(ctx context.Context) error {
+	buckets, err := tc.listBuckets(ctx)
```
maurycy, line 240: @dmerrill6 

I might be wrong but my understanding is that `listBuckets()` return `nil, err` `if tc.isRunning == false || tc.isInitialized == false` and this might be exactly the case here `since tc.isInitialized = true` is only done later.

maurycy (commented):
@dmerrill6 

Some quick, probably unsubstantiated, comments.

perfectmak (approved):

core/textile/client.go
dmerrill6: I updated `ListBuckets` to do the `requiresRunning` check, but `initialize` is calling `listBuckets` instead, which doesn't check for these conditions.

The idea behind this is that if `ListBuckets` (the public method) is called from outside and the client is not running, we return an error.

core/textile/client.go
dmerrill6: I tested this a bunch and it's not flooding because the healthcheck stops after failing to get a public key, thus not even getting to do anything at the network. If there are keys and it passes this barrier, then it goes to `getHubCtx` which returns a cached data from the store so there shouldn't be flooding either. In any case I think we can test this and revisit if we find any issues.

core/textile/client.go
dmerrill6: Fixed in 6a1fced

app/app.go
dmerrill6: This was a leftover from the merge with master. Fixed in 6a1fced.

core/textile/client.go
```diff
@@ -313,3 +334,30 @@ func (tc *textileClient) getThreadContext(parentCtx context.Context, threadName
...
+
+// Checks for connection and initialization needs.
+func (tc *textileClient) healthcheck(ctx context.Context) {
+	log.Debug("Textile Client healthcheck... Start.")
+
+	if tc.isInitialized == false {
+		tc.initialize(ctx)
+	}
+
+	tc.ConnectToHub(ctx)
```
jsonsivar, line 346: Should we wrap this in a `tc.isConnectedToHub == false` clause?

studna (approved):

core/textile/client.go
```diff
@@ -238,14 +250,27 @@ func (tc *textileClient) Start(ctx context.Context, cfg config.Config) error {
...
 	}
 
+	tc.isInitialized = true
+	return nil
+}
+
+// Starts a Textile Client and also initializes default resources for it like a key pair and default bucket.
+// Then leaves the process running to attempt to connect or to initialize if it's not already initialized
+func (tc *textileClient) Start(ctx context.Context, cfg config.Config) error {
+	// Create key pair if not present
```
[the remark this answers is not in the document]
dmerrill6: Fixed in 5660fd4

core/textile/client.go
```diff
@@ -137,23 +162,36 @@ func (tc *textileClient) SetUc(uc UsersClient) {
...
 		log.Error("Could not connect to Textile Hub. Starting in offline mode.", err)
+		return err
 	}
 
+	wasConnectedToHub := tc.isConnectedToHub
+
 	tc.isConnectedToHub = true
 
 	// setup mailbox
 	mid, err := tc.uc.SetupMailbox(hubctx)
```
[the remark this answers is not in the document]
dmerrill6: Fixed in 5660fd4

core/textile/client.go
dmerrill6: Updated `ConnectToHub` to better reflect what it's doing right now. So now it's called `checkHubConnection` and it will get the token from the store, refresh it if necessary, and connect to mailbox if necessary. If any of these steps fail, `tc.isConnectedToHub` will be set to `false`.

core/keychain/keychain.go
```diff
@@ -204,18 +204,15 @@ func (kc *keychain) DeleteKeypair() error {
 		return err
 	}
 
-	err = ring.Remove(PrivateKeyStoreKey)
-
-	err2 := kc.st.Remove([]byte(PublicKeyStoreKey))
+	// Note: currently ignoring error on keychain removal because it's failing randomly.
+	// Use GenerateKeyPair with override option instead.
+	ring.Remove(PrivateKeyStoreKey)
```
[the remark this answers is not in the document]
dmerrill6: Created a card here https://app.clubhouse.io/terminalsystems/story/18171/error-when-deleting-key-ring-nonblocking

core/textile/client.go
```diff
@@ -125,7 +129,28 @@ func (tc *textileClient) start(ctx context.Context, cfg config.Config) error {
 	tc.ConnectToHub(ctx)
```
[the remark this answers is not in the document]
dmerrill6: Removed in 5660fd4

core/textile/client.go
maurycy: @dmerrill6 

You are right! My bad!

core/textile/client.go
maurycy: @dmerrill6 

Makes sense, thank you!

PR #142, opened 2020-09-01
Set ulimit on start

dmerrill6:
Was trying to fix the `too many open files` error. Followed some leads and saw this https://docs.mongodb.com/manual/reference/ulimit/#:~:text=Most%20UNIX%2Dlike%20operating%20systems,using%20too%20many%20system%20resources and this https://github.com/ipfs/go-ipfs/issues/4589#issuecomment-358746222

After setting ulimit to a very high value, the problem seemed to be fixed. In this PR I am setting the ulimit value to the maximum possible on startup. Need further testing in other machines.

[ch18171]

### DIFF
--- a/cmd/space-daemon/main.go
+++ b/cmd/space-daemon/main.go
@@ -4,19 +4,18 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"math"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
 	"runtime"
 	"runtime/pprof"
-	"syscall"
 
 	"github.com/FleekHQ/space-daemon/log"
 
 	"github.com/FleekHQ/space-daemon/app"
 	"github.com/FleekHQ/space-daemon/config"
 	"github.com/FleekHQ/space-daemon/core/env"
+	"github.com/FleekHQ/space-daemon/core/util/rlimit"
 )
 
 var (
@@ -96,7 +95,7 @@ func main() {
 	// setup context
 	ctx := context.Background()
 
-	setULimit()
+	rlimit.SetRLimit()
 
 	spaceApp := app.New(cfg, env)
 	// this blocks and returns on exist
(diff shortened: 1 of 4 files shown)


cmd/space-daemon/main.go
```diff
@@ -109,6 +113,33 @@ func main() {
...
+	if err != nil {
+		log.Error("Error Getting Rlimit. Please run `ulimit -n 1000` from a privileged user to avoid issues when running the space daemon.", err)
+		return
+	}
+	log.Debug(fmt.Sprintf("Got Rlimit: Cur: %d, Max: %d", rLimit.Cur, rLimit.Max))
+
+	// Max allowed value is 10240 even when rLimit.Max can go beyond that
+	rLimit.Cur = uint64(math.Min(10240, float64(rLimit.Max)))
+
+	err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit)
```
jsonsivar, line 134: Is this on a per-process basis? Just trying to think if this impacts users with machines with low mem or a leak happens. But we can probably deal with that if it happens.

Another thing was that the `lsof` dump from @giankotarola had textile having about 160 files open re: it's `4001` port. That's weird because @perfectmak put in a fix to limit textile to 100 by default so not sure how that could be over.  Just putting that out there but maybe not worth it now to look further and we can come back to this later.

maurycy: I'm a bit worried about different platforms.

This is how https://github.com/ipfs/go-ipfs handles `ulimit`:

https://github.com/ipfs/go-ipfs/tree/master/cmd/ipfs/util
https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/util/ulimit.go#L46
https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/util/ulimit_unix.go
https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/util/ulimit_windows.go
https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/util/ulimit_freebsd.go

maurycy (approved):
LGTM. 🚀

cmd/space-daemon/main.go
dmerrill6: Noticed this when the windows build failed. Sent a fix in 1c0e374. Agree we can test with this new settings and go from there.

perfectmak (approved):
LGTM!

PR #146, opened 2020-09-02
Add address derivation function

dmerrill6:
# Change log

- Add address derivation function and display the address when initiating textile client


jsonsivar (approved):
LGTM :+1:

maurycy (approved):

PR #147, opened 2020-09-03
Fix/Sharing: added full path to request, added more placeholders.

jsonsivar:
### Change Log
* Added `FullPath` to share request so they can be passed down
* Some additional refactoring/placeholders in lieu of full integration

### TODO
* Get mirror bucket thread ID and set in paths
* Get key for file


core/space/services/services_sharing.go
```diff
@@ -225,12 +225,30 @@ func (s *Space) OpenSharedFile(ctx context.Context, hash, password, filename str
...
+
+	for _, path := range paths {
+		if path.DbId == "" {
+			// TODO: fetch dbid of mirror bucket and set
+			// b, err := s.tc.GetDefaultBucket(ctx)
+			// if err != nil {
+			// 	return err
+			// }
+
+			// bs, err := s.tc.GetBucketSchema(ctx, b.Slug())
```
jsonsivar, line 242: @maurycy just wanted to check if there is any function like this coming with your PR before I go ahead and create one (i.e., external function to get the schema given the slug).

core/space/services/services_sharing.go
```diff
@@ -225,12 +225,30 @@ func (s *Space) OpenSharedFile(ctx context.Context, hash, password, filename str
...
+			// }
+			// path.DbId == bs.dbId
+		}
+	}
+
 	for _, pk := range pubkeys {
+
 		d := &domain.Invitation{
 			Paths: paths,
 			// Key: TODO - get from keys thread for each file
```
jsonsivar, line 254: @perfectmak could you direct me to somewhere to be able to fetch the keys given a path? I can create one if there isn't one coming with your work.

core/textile/sharing.go
```diff
@@ -3,22 +3,24 @@ package textile
...
 	// for _, pth := range paths {
-	// 	var roles map[string]buckets.Role
-	// 	for _, pk := range pubkeys {
-	// 		roles[pk] = buckets.Role.Writer
-	// 	}
-	// 	err := tc.EditPathAccessRoles(ctx, key, pth, roles)
-	// 	if err != nil {
-	// 		return err
-	// 	}
+	//   ctx = tc.GetRemoteBucketContext(ctx, pth.DbId, pth.Bucket)
```
jsonsivar, line 15: @maurycy this is another one I wanted to double check if there is overlap before I go ahead and create one.

core/space/services/services_sharing.go
maurycy: @jsonsivar 

For `BucketSchema` (not `textile.Bucket`) there is:

```
func (tc *textileClient) findBucketInCollection(ctx context.Context, bucketSlug string) (*BucketSchema, error) {
```

Does that work for you?

dmerrill6 (approved):
LGTM

gpuente (approved):
LGTM

core/space/services/services_sharing.go
jsonsivar: Yes it does, I made that function public for now and used that thank you.

core/space/services/services_sharing.go
perfectmak: I added the encryption key to the bucketschema here: https://github.com/FleekHQ/space-daemon/pull/123/files#diff-8ea426b2ef49623ad0f07dd826f93169R21

It is bucket level and works for all paths under the parent bucket.

core/textile/bucket_factory.go
```diff
@@ -53,6 +53,22 @@ func getThreadName(userPubKey []byte, bucketSlug string) string {
...
 
+func (tc *textileClient) getRemoteBucketContext(ctx context.Context, sDbID string, bucketSlug string) (context.Context, error) {
+	log.Debug("getRemoteBucketContext: Getting remote bucket context")
+
+	dbID, err := utils.ParseDbIDFromString(sDbID)
+	if err != nil {
+		log.Error("Error casting thread id", err)
+		return nil, err
+	}
+	ctx, err = tc.getThreadContext(ctx, bucketSlug, *dbID, true)
```
dmerrill6, line 64: Feels a bit weird that this has a different behaviour than `getBucketContext`, in the sense that the latter creates a bucket if it's not already present and adds it to the metathread collection. Should we instead add a flag to that method to tell it if it should use the hub context or not when calling `getThreadContext`?

jsonsivar: That works too.  I noticed that difference but wanted to keep this one light instead of all the logic to create it if it doesn't exist because we won't have this one in the collection the same way either so that piece needs to account for this particular bucket not being in the collections.  Maybe we can go with this design to keep it clean:
- Rename my new function to an inner `getBucketContext` that just simply tries to create the context with the `hub` `bool` like you said (I could use this for my use case)
- Turn the existing getBucketContext to something like getOrCreateBucketContext that calls the inner one from 1

Let me know what you think

dmerrill6: Sounds good to me. We should update the `getOrCreateBucketContext` to internally use your new function then instead of calling `tc.getThreadContext` directly.

jsonsivar: Fixed in [b842036](https://github.com/FleekHQ/space-daemon/pull/147/commits/b8420364f85811c6db50e8bb634d4e1d948b9a3f).

dmerrill6 (approved):
LGTM

PR #149, opened 2020-09-03
Hotfix/sharing: just like dbid, this makes bucket optional for default buckets

jsonsivar:
This is so the FE doesn't have to send the repeated "personal" bucket name for each element in the paths on input (but we still need to have the field on a per path basis because for shared-with-me items they can vary file to file).


gpuente (approved):
LGTM

dmerrill6 (approved):
LGTM

PR #150, opened 2020-09-03
Refactor collections and "Shared with me" thread

dmerrill6:
Tagging everyone so that we can merge this before including new developments to textile client, otherwise we will have lots of conflicts 😅 

# Change log

- Refactor textile client to decouple collections (now called models) from the client itself

- Refactor the models so that each model lives in its own file

- Add new model for shared with me (now called received files). TODO: Wire it up once we add the handler for receiving shares.

[ch18197]


maurycy:
😨

core/textile/model/buckets.go
```diff
@@ -0,0 +1,158 @@
...
+	id := res[0]
+	return &BucketSchema{
+		Slug:   newInstance.Slug,
+		ID:     core.InstanceID(id),
+		DbID:   newInstance.DbID,
+		Backup: newInstance.Backup,
+	}, nil
+}
+
+func (m *model) UpsertBucket(ctx context.Context, bucketSlug, dbID string) (*BucketSchema, error) {
```
jsonsivar, line 64: Since there are more fields coming on this schema from the mirror bucket work, maybe we make this take in a bucket schema object so you can set any field.  cc @maurycy

dmerrill6:
> 😨

If you already went off the current master branch no worries, we can merge yours first and then I rebase.

maurycy:
@dmerrill6 

Thank you. I'm waiting for the final review from @jsonsivar (I was confused by all the `context` transformations.) If there are no major holes in my pull request, I'd prefer to merge first.

jsonsivar (approved):
LGTM

core/textile/model/buckets.go
maurycy: @jsonsivar 

Yes, I love the `model` separation. We even discussed this today, ie, moving different collections into different files in the `collection` package.

When it comes to the `func` signatures, I have mixed feelings. I have created a separate `struct` for an upsert myself and I don't like exposing, ie, returning, the internal `*Schema` (the one with `json`), but not really sure if we should multiplicate the `struct` for every CRUD.

maurycy (approved):
LGTM.

perfectmak (approved):

PR #151, opened 2020-09-04
Add mailbox to init flow

dmerrill6:


jsonsivar (approved):
LGTM

core/textile/client.go
```diff
@@ -197,6 +199,12 @@ func (tc *textileClient) checkHubConnection(ctx context.Context) error {
 			return err
 		}
 		tc.mb = mailbox
+
+		if err := tc.listenForMessages(hubctx); err != nil {
+			tc.isConnectedToHub = false
```
maurycy, line 204: @dmerrill6 

Do you think we need any locking here?

maurycy (approved):
LGTM. 🚀

core/textile/client.go
dmerrill6: Will merge to unblock the FE and we can iterate if we find issues

maurycy: @dmerrill6 

Yes, let's discuss this later.

PR #152, opened 2020-09-04
mirror bucket and backup files

dmerrill6:
Rebase of #139


dmerrill6:
Brought the changes in #139 branching from current master. 

Still need to review myself as this was just making it compile but not thinking thoroughly if something could be improved.

core/space/services/services_fs.go
```diff
@@ -492,6 +493,22 @@ func (s *Space) addFile(ctx context.Context, sourcePath string, targetPath strin
 		return domain.AddItemResult{}, err
 	}
 
+	if s.tc.IsBucketBackup(ctx, b.Slug()) && !s.tc.IsMirrorFile(ctx, targetPath, b.Slug()) {
```
dmerrill6, line 496: This logic might be better off inside textile client rather than the service. My thinking is that this pretty much only accesses textile client functions (`UploadFileToHub`, `MarkMirrorFileBackup`), but these ops seem internal concerns of the textile client rather than stuff that should be called from outside.

Maybe we can expose a single `tc.AddFile` that internally checks for these flags (`IsBucketBackup`) and then chooses wether to mirror or not. That way all these mirror functions can remain private.

core/textile/bucket_factory.go
```diff
@@ -200,11 +201,23 @@ func (tc *textileClient) createBucket(ctx context.Context, bucketSlug string) (B
...
+		return nil, err
+	}
+
+	mirrorSchema, err := tc.createMirrorBucket(ctx, *schema)
 	if err != nil {
 		return nil, err
 	}
 
+	if mirrorSchema != nil {
+		_, err = m.CreateMirrorBucket(ctx, bucketSlug, mirrorSchema)
```
dmerrill6, line 215: It is confusing that this function is calling both `model.CreateMirrorBucket` and `tc.CreateMirrorBucket`. Maybe the collection creation can be taken care of inside tc.createMirrorBucket`?

core/textile/model/mirror_file.go
```diff
@@ -0,0 +1,142 @@
...
+type MirrorBucketSchema struct {
+	RemoteDbID string `json:"remoteDbId"`
+	HubAddr    string `json:"HubAddr"`
+}
+
+const mirrorFileModelName = "MirrorFile"
+
+var errMirrorFileNotFound = errors.New("Mirror file not found")
+
+func (m *model) CreateMirrorBucket(ctx context.Context, bucketSlug string, mirrorBucket *MirrorBucketSchema) (*BucketSchema, error) {
```
dmerrill6, line 35: Maybe this can be moved to `model/buckets`. Not sure where it fits best.

core/textile/textile.go
```diff
@@ -75,6 +75,10 @@ type Client interface {
 	ShareFilesViaPublicKey(ctx context.Context, paths []domain.FullPath, pubkeys []crypto.PubKey) error
 	RemoveKeys() error
 	AttachMailboxNotifier(notif GrpcMailboxNotifier)
+	IsBucketBackup(ctx context.Context, bucketSlug string) bool
+	IsMirrorFile(ctx context.Context, path, bucketSlug string) bool
+	UploadFileToHub(ctx context.Context, b Bucket, path string, reader io.Reader) (result path.Resolved, root path.Path, err error)
+	MarkMirrorFileBackup(ctx context.Context, path, bucketSlug string) (*domain.MirrorFile, error)
```
dmerrill6, line 81: We should be able to remove these to avoid polluting the public interface, or maybe replace with a single method

core/textile/model/mirror_file.go
```diff
@@ -0,0 +1,142 @@
...
+	err = m.threads.Save(metaCtx, *metaDbID, bucketModelName, instances)
+	if err != nil {
+		return nil, err
+	}
+
+	return bucket, nil
+}
+
+// Finds the metadata of a file that has been shared to the user
+func (m *model) FindMirrorFileByPathAndBucketSlug(ctx context.Context, path, bucketSlug string) (*MirrorFileSchema, error) {
```
jsonsivar, line 60: `bucketSlug` is not used here. I think it doesn't cause any issues since the bucket slug is always `personal` but maybe we can add that to the `Find` query just in case.

core/textile/model/mirror_file.go
```diff
@@ -0,0 +1,142 @@
...
+	log.Debug("stored mirror file with dbid " + newInstance.DbID)
+
+	id := res[0]
+	return &MirrorFileSchema{
+		Path:       newInstance.Path,
+		BucketSlug: newInstance.BucketSlug,
+		Backup:     newInstance.Backup,
+		Shared:     newInstance.Shared,
+		ID:         core.InstanceID(id),
+		DbID:       newInstance.DbID,
```
jsonsivar, line 117: Is this set?

jsonsivar (commented):
Left a few comments. Most of them are minor or questions but there are 2 blocker ones I think we need to address:

1. Using the remote buckets `.key` value when uploading a file
2. Creating a thread id for the remote bucket during creation.

jsonsivar:
Just wanted to add one more note to this, we should use the new `SecureBucketClient` when adding the file to the hub.

jsonsivar (approved):
LGTM now.  We updated the critical comments but for the remainder I guess we can incur the debt and do them later?

dmerrill6:
> LGTM now. We updated the critical comments but for the remainder I guess we can incur the debt and do them later?

Debt accepted, merging after CI passes.

maurycy:
@dmerrill6 
@jsonsivar 

I'm fine with the debt. I will add the necessary comments.

dmerrill6:
> @dmerrill6
> @jsonsivar
> 
> I'm fine with the debt. I will add the necessary comments. Thank you!

Let me know when the comments are added so I can merge.

PR #155, opened 2020-09-04
recentlyShared

maurycy:


dmerrill6 (commented):
This needs to be stored in a thread, otherwise it will get cleared after turning off the daemon

dmerrill6 (approved):
LGTM!

PR #159, opened 2020-09-05
Hotfix/mirror: fix create mirror on init

jsonsivar:
### Change Log
* Create mirror thread was missing hub ctx so fixed that
* Moved `initialize`to after hub connection start to avoid race condition

NOTE: this means that a hub connection is required to init for now.  I think this is okay to release and then we can have a better online vs offline state management work started asap in parallel (i.e., what happens if they are offline during init? and then what happens if they come back online post init and vice versa).

### DIFF
--- a/core/textile/client.go
+++ b/core/textile/client.go
@@ -351,11 +351,11 @@ func (tc *textileClient) IsRunning() bool {
 func (tc *textileClient) healthcheck(ctx context.Context) {
 	log.Debug("Textile Client healthcheck... Start.")
 
+	tc.checkHubConnection(ctx)
+
 	if tc.isInitialized == false {
 		tc.initialize(ctx)
 	}
-
-	tc.checkHubConnection(ctx)
 
 	switch {
 	case tc.isInitialized == false:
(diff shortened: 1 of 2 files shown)


core/textile/client.go
```diff
@@ -351,12 +351,12 @@ func (tc *textileClient) IsRunning() bool {
 func (tc *textileClient) healthcheck(ctx context.Context) {
 	log.Debug("Textile Client healthcheck... Start.")
 
+	tc.checkHubConnection(ctx)
```
maurycy, line 354: @jsonsivar 

What do you think about documenting the trade-off?

maurycy (commented):
LGTM. 🏇🏻

maurycy (approved):

core/textile/client.go
jsonsivar: Good idea, will add a comment in the code before merging.

dmerrill6: Do we need to return if we get error here? otherwise if you are offline, that call will fail and then the initialize will still not create the mirror bucket but it will create the local one thus not ever creating the mirror one again

jsonsivar: Fixed in [9e3dec9](https://github.com/FleekHQ/space-daemon/pull/159/commits/9e3dec9634c08fe5cd7e71f3ba487f353dc63000). 

I set it so that the init only happens is there is no error on the hub connection so if you are offline it will keep trying to init until you are online again.  Not ideal but we can add a better fix after further discussion so we can support offline/online better as per the comment.

dmerrill6 (approved):

PR #160, opened 2020-09-07
Hotfix/mirror: fixes a few bugs in the mirror bucket and mirror file creation

jsonsivar:
### Change Log
- `UploadFileToHub` was calling get bucket context with some params swapped
- DbId casting issue
- Were not saving remote bucket key which was needed for lookup
- `FindMirrorFileByPathAndBucketSlug` - don't throw error on not found, instead return empty object (so it doesn't break creation)
- Use `cfg` value instead of OS environ directly for mirror hub addr

### DIFF
--- a/core/textile/mirror.go
+++ b/core/textile/mirror.go
@@ -3,8 +3,8 @@ package textile
 import (
 	"context"
 	"io"
-	"os"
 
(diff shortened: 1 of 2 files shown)


maurycy (approved):

maurycy:
Great finds, thank you! 🚋

PR #161, opened 2020-09-08
Hotfix/sharing: some more bugs with sharing and mirror file creation

jsonsivar:
### Change Log
* For creating mirror file, use `targetBucketPath` instead of `targetPath` so we don't pass in emtpy string for root files
* Pass `paths` in as a pointer so `ShareFilesViaPublicKey` can modify it for enhancing data
* Another dbid casting error
* For `FindMirrorFileByPathAndBucketSlug` don't return empty struct so nil could be checked for when trying to create
* Handle file already exists for mirror file collection insert

### DIFF
--- a/core/space/services/services_sharing.go
+++ b/core/space/services/services_sharing.go
@@ -252,8 +252,6 @@ func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths []domain.FullP
 				return err
 			}
 
-			log.Info("incoming dbid nil, updating")
-			log.Info("using new remote db id: " + bs.RemoteDbID)
 			ep.DbId = bs.RemoteDbID
 		}
 
--- a/core/space/services/services_sharing.go
+++ b/core/space/services/services_sharing.go
@@ -13,12 +13,12 @@ import (
 
 	"github.com/FleekHQ/space-daemon/log"
 	crypto "github.com/libp2p/go-libp2p-crypto"
+	"github.com/pkg/errors"
 
 	"github.com/textileio/dcrypto"
 
-	"github.com/pkg/errors"
-
 	"github.com/FleekHQ/space-daemon/core/space/domain"
+	t "github.com/FleekHQ/space-daemon/core/textile"
 	"github.com/ipfs/go-cid"
 )
 
@@ -226,11 +226,12 @@ func (s *Space) OpenSharedFile(ctx context.Context, hash, password, filename str
 	}, nil
 }
 
-func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths []domain.FullPath, pubkeys []crypto.PubKey) error {
+func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths *[]domain.FullPath, pubkeys []crypto.PubKey) error {
 	m := s.tc.GetModel()
 
-	var enckeys [][]byte
-	for i, path := range paths {
+	enckeys := make([][]byte, len(*paths))
+	for i, path := range *paths {
+
 		// this handles personal bucket since for shared-with-me files
 		// the dbid will be preset
 		if path.DbId == "" {
@@ -239,14 +240,14 @@ func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths []domain.FullP
 				return err
 			}
 
-			bs, err := m.FindBucket(ctx, b.GetData().Name)
+			bs, err := m.FindBucket(ctx, b.Slug())
 			if err != nil {
 				return err
 			}
-			path.DbId = bs.RemoteBucketKey
+			(*paths)[i].DbId = bs.RemoteDbID
 		}
 
-		if path.Bucket == "" {
+		if path.Bucket == "" || path.Bucket == t.GetDefaultBucketSlug() {
 			b, err := s.tc.GetDefaultBucket(ctx)
 			if err != nil {
 				return err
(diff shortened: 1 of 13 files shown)


maurycy (approved):

core/space/services/services_sharing.go
```diff
@@ -226,11 +226,12 @@ func (s *Space) OpenSharedFile(ctx context.Context, hash, password, filename str
 	}, nil
 }
 
-func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths []domain.FullPath, pubkeys []crypto.PubKey) error {
+func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths *[]domain.FullPath, pubkeys []crypto.PubKey) error {
```
perfectmak, line 229: Maybe add a comment on this function that `paths` would be modified. Personally, I think it would have been better if the function returned the modified `paths` instead of mutating the input value, but eitherways a comment should help make it more apparent.

perfectmak (approved):
LGTM! I left a comment

dmerrill6 (approved):

studna (approved):

core/textile/model/mirror_file.go
```diff
@@ -89,11 +90,15 @@ func (m *model) CreateMirrorFile(ctx context.Context, mirrorFile *domain.MirrorF
...
 	}
 
-	_, err = m.FindMirrorFileByPathAndBucketSlug(ctx, mirrorFile.Path, mirrorFile.BucketSlug)
+	mf, err := m.FindMirrorFileByPathAndBucketSlug(ctx, mirrorFile.Path, mirrorFile.BucketSlug)
 	if err != nil {
 		return nil, err
 	}
 
+	if mf != nil {
+		return nil, errMirrorFileAlreadyExists
```
maurycy, line 99: @jsonsivar 

I wonder: should we return `err` or just `return mf, nil`?

core/textile/bucket_factory.go
```diff
@@ -200,7 +200,7 @@ func (tc *textileClient) createBucket(ctx context.Context, bucketSlug string) (B
 
 	// We store the bucket in a meta thread so that we can later fetch a list of all buckets
 	log.Debug("Bucket " + bucketSlug + " created. Storing metadata.")
-	schema, err := m.CreateBucket(ctx, bucketSlug, dbID.String())
+	schema, err := m.CreateBucket(ctx, bucketSlug, utils.CastDbIDToString(*dbID))
```
maurycy, line 203: 🤦🏻 

I started with `utils.CastDbIDToString()` and later changed it to `dbID.String()` because it looked more intuitive.

core/space/services/services_sharing.go
jsonsivar: Fixed in [26a4678](https://github.com/FleekHQ/space-daemon/pull/161/commits/26a4678264f4dc0e08047c1693559ae8308a5772).

core/textile/model/mirror_file.go
jsonsivar: Will leave this as-is for now and can update later since it's working at least.  If we need a create or get type usage if this later we can always update.

core/space/services/services_sharing.go
```diff
@@ -244,10 +251,13 @@ func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths *[]domain.Full
 			if err != nil {
 				return err
 			}
-			(*paths)[i].DbId = bs.RemoteDbID
+
+			log.Info("incoming dbid nil, updating")
+			log.Info("using new remote db id: " + bs.RemoteDbID)
```
dmerrill6: Should these logs be Debug? Seems like they will always be printing when sharing shared-with-me files.

core/space/services/services_sharing.go
```diff
@@ -244,10 +251,13 @@ func (s *Space) ShareFilesViaPublicKey(ctx context.Context, paths *[]domain.Full
...
+			ep.DbId = bs.RemoteDbID
 		}
 
-		if path.Bucket == "" || path.Bucket == t.GetDefaultBucketSlug() {
+		if ep.Bucket == "" || ep.Bucket == t.GetDefaultBucketSlug() {
```
dmerrill6, line 258: Do shared with me files have `ep.Bucket == ""`? Because if that's the case it might unintendedly fall into this block for shared-with-me files

dmerrill6 (approved):
LGTM, left some minor questions

core/space/services/services_sharing.go
jsonsivar: These were removed in the next commit :+1:

perfectmak (approved):

core/space/services/services_sharing.go
jsonsivar: So shared with me files will always have a bucket name. But the weird part is that the mirror bucket name is also called `personal` so it would erroneous fall into the second clause.  I made both clauses so that it would default to the personal bucket if it was empty or if `personal` was supplied explicitly. For now I'm thinking to keep it clean we can change the name of the mirrored buckets to something like `shared` so it doesn't fall into this.

jsonsivar: @maurycy do you see any issues with this idea? Like if I used `personal_mirror` for the mirror bucket names. As long as we are not referring the mirror bucket name hardcoded as `personal` elsewhere that should be fine.

maurycy: @jsonsivar 

Not sure if I understand. Do you mean adding a prefix, eg: `shared_`, `personal_`?

jsonsivar: Fixed in [1742bb0](https://github.com/FleekHQ/space-daemon/pull/161/commits/1742bb01750329eae1ca85f136fa508338aef156).  Mirror buckets now use a different name so it wouldn't fall into this slot unless it was actually a personal bucket (or empty in which case it is personal).

PR #164, opened 2020-09-08
Open shared files

dmerrill6:
# Change log

- Update `GetBucket` handlers to allow receiving a `dbID` to connect to a hub thread.

- Update `OpenFile` to allow opening shared files.

- Fix some concurrency bugs when doing the jwt authorization flow.

~Needs further testing after sharing bugs are solved.~ Tested and it's now working after solving some bugs on the storage of shared files.

[ch18236]


core/textile/hub/hub_auth.go
```diff
@@ -67,16 +68,18 @@ type HubAuth interface {
...
 
 func New(st store.Store, kc keychain.Keychain, cfg config.Config) *hub {
 	return &hub{
-		st:  st,
-		kc:  kc,
-		cfg: cfg,
+		st:               st,
+		kc:               kc,
+		cfg:              cfg,
+		fetchTokensMutex: &sync.Mutex{},
```
jsonsivar, line 82: This is already in alpha branch right? Looks harmless but flagging in case some merge issue is there.

core/textile/mirror.go
```diff
@@ -52,7 +52,12 @@ func (tc *textileClient) UploadFileToHub(ctx context.Context, b Bucket, path str
 		return nil, nil, err
 	}
 
-	return tc.hb.PushPath(hubCtx, bucket.RemoteBucketKey, path, reader)
+	bucketsClient := NewSecureBucketsClient(
```
jsonsivar, line 55: :pray:

jsonsivar (approved):
LGTM :rocket:

core/textile/hub/hub_auth.go
```diff
@@ -229,6 +232,8 @@ func (h *hub) getTokensThroughChallenge(ctx context.Context) (*inMessageTokenVal
 }
 
 func (h *hub) GetTokensWithCache(ctx context.Context) (*AuthTokens, error) {
+	h.fetchTokensMutex.Lock()
```
maurycy, line 235: @dmerrill6 

🙇🏻

maurycy (approved):
LGTM. 🪂

core/textile/hub/hub_auth.go
dmerrill6: Added this additional lock apart from the one in master given that the hub is accessed directly from the space service when doing `GetSessionTokens`, which sometimes triggered both healthcheck and that to go off at the same time

PR #165, opened 2020-09-09
Hotfix/sharing-accept: hotfixes for accepting shares and some related fixes

jsonsivar:
### Change Log
* Properly pass encryption keys via share message
* Add missing inviter and invitee pub key in message
* Handle parsing invitation reply on get notifications
* Don't throw on unknown message type since technically users can get any mail outside of the app too and we don't want to break UX on that
* Only send invitation accept message if receiver accepts (before it was sending on reject too)


core/space/services/services_sharing.go
```diff
@@ -338,38 +353,38 @@ func (s *Space) HandleSharedFilesInvitation(ctx context.Context, invitationId st
...
+		})
 
+		if err != nil {
+			log.Error("error encoding invitation response", err)
+			return errFailedToNotifyInviter
+		}
+
+		_, err = s.tc.SendMessage(ctx, invitersPk, message)
+	} else {
+		invitation, err = s.tc.RejectSharedFilesInvitation(ctx, invitation)
```
perfectmak, line 382: I see we don't want to notify inviter that his invite was rejected?

perfectmak (approved):

core/space/services/services_sharing.go
jsonsivar: Yeah I was thinking like when we get Github invites or docs I don't remember usually getting invites for rejections.  But spoke with Harrison on this and we can leave it like this and see if we want reject notifications when we do the polishing and tweak since it's a small change to do it on rejects as well.

PR #166, opened 2020-09-09
Hotfix/receivedfiles: handle error and formatting

jsonsivar:
### Change Log
* Don't throw if members query throws (this is a temp fix and can be removed once we fix on textile side)
* Fix members list not showing properly
* Fix formatting for size and date fields


perfectmak (approved):

maurycy (approved):
LGTM. 🚂

PR #168, opened 2020-09-09
Hotfix/notifs: dont send extra path item

jsonsivar:
This bug was sending an empty path item before any real path items.


maurycy (approved):
LGTM. 🤠

PR #169, opened 2020-09-10
Upload public shared files to hub

perfectmak:
**CHANGELOG**
- For every user create a public thread and each thread has a `personal_public` bucket where publicly shared files will be uploaded to.
- Decrypting publicly shared files are downloaded from Hub Gateway.
- Add a new main flag `textilehubgatewayurl` and environment variable `TXL_HUB_GATEWAY_URL` for it. Currently it defaults to "https://hub-dev.space.storage" if no value is passed.


core/space/services/services_sharing.go
```diff
@@ -86,18 +87,25 @@ func (s *Space) GenerateFileSharingLink(
...
-	fileUploadResult := s.ic.AddItem(ctx, sharedContent)
-	if err := fileUploadResult.Error; err != nil {
-		return EmptyFileSharingInfo, errors.Wrap(err, "encrypted file upload failed")
+	b, err := s.tc.GetPublicShareBucket(ctx)
+	if err != nil {
+		return EmptyFileSharingInfo, errors.Wrap(err, "failed to get public files bucket")
 	}
-	encryptedFileHash := fileUploadResult.Resolved.Cid().String()
+
+	uploadResult, _, err := b.UploadFile(ctx, fmt.Sprintf("%s-%d", fileName, time.Now().UnixNano()), sharedContent)
```
perfectmak, line 103: I'm currently getting an rpc error at this line:
```
 rpc error: code = Unknown desc = instance not found
```
I think it has to do with something not yet created on the hubs thread, other this, this is ready for review.

core/textile/bucket_factory.go
```diff
@@ -231,10 +231,6 @@ func (tc *textileClient) createBucket(ctx context.Context, bucketSlug string) (B
 		return b, nil
 	}
 
-	if err != nil {
```
maurycy, line 234: @perfectmak 

Why?

core/textile/utils/utils.go
```diff
@@ -36,7 +36,7 @@ func ParseDbIDFromString(dbID string) (*thread.ID, error) {
 
 type DeterministicThreadVariant string
 
-var (
+const (
```
maurycy, line 39: @perfectmak 

👍🏻

maurycy (approved):
LGTM. 🎸

core/ipfs/utils.go
```diff
@@ -66,4 +67,19 @@ func GetFileHash(r io.Reader) (string, error) {
...
+func DownloadIpfsItem(ctx context.Context, gatewayUrl string, cid cid.Cid) (io.ReadCloser, error) {
+	url := fmt.Sprintf("%s/ipfs/%s", gatewayUrl, cid.String())
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
 
+	client := http.Client{}
+
+	resp, err := client.Do(req.WithContext(ctx))
```
maurycy, line 79: @perfectmak 

What do you think about a timeout?

jsonsivar (approved):

jsonsivar:
Only thing is to add the new env var in `.github/workflows/release.yml` and `.goreleaser.yml`.

core/ipfs/utils.go
perfectmak: Yeah, I initially added a 5 mins time-out, but realized that sometimes people might shared large files and it might take a while to download completely, so just resorted to make the request cancelable via the context. So if the request is cancelled, ideally the download should cancel as well.

core/textile/bucket_factory.go
perfectmak: It was useless... `err` before this statements is nil, so I just deleted it as redundant code

maurycy: @perfectmak 

True that! Confused by GitHub UI. 🙇🏻

core/textile/public.go
```diff
@@ -0,0 +1,136 @@
...
+	err = tc.store.Set([]byte(publicShareThreadStoreKey), []byte(dbId))
+	if err != nil {
+		return thread.Undef, errors.Wrap(err, "failed to persist public share thread")
+	}
+
+	return dbId, nil
+}
+
+// DownloadPublicGatewayItem download a cid content from the hubs public gateway
+func (tc *textileClient) DownloadPublicGatewayItem(ctx context.Context, cid cid.Cid) (io.ReadCloser, error) {
```
maurycy, line 133: @perfectmak 

What do you think about rate limiting here? We effectively let people to use us as an IPFS gateway proxy. :-)

cc @jsonsivar

perfectmak:
> Only thing is to add the new env var in .github/workflows/release.yml and .goreleaser.yml.

@jsonsivar this has been added here https://github.com/FleekHQ/space-daemon/pull/169/commits/862c3aa9a975f5eef3d110ecf760520a7628602b

core/textile/public.go
perfectmak: Hmmm... I'm not sure how that could be a problem as the gateway url is actually public and anyone can access it outside of the daemon. And this `DownloadPublicGatewayItem()` is not directly exposed in the daemon and it currently only used via the open public shared files rpc call and if an abitrary cid not generated by the daemon user make a request through that rpc, it would get an incorrect password error.

maurycy: @perfectmak 

Gotcha! Thank you for the explanation.

jsonsivar (approved):

PR #171, opened 2020-09-10
Make key generation / restore ops wait for initialization of textile

dmerrill6:
# Change log

- Solve bug ch18273 where the clientside was requesting stuff from textile client before it was ready (right after account seeding).

[ch18273]


jsonsivar (approved):
LGTM.  Mocks need to be updated but other than that looks good to go.

maurycy (approved):
LGTM. 🐦

perfectmak (approved):

PR #173, opened 2020-09-10
listDirAtPath: expose members

maurycy:


core/space/domain/domain.go
```diff
@@ -154,5 +155,5 @@ type SharedDirEntry struct {
 	DbID   string
 	Bucket string
 	FileInfo
-	Members []Member
+	Members []Member // XXX: it is duplicated from FileInfo
```
jsonsivar, line 158: Okay maybe we can remove this then since it's included in FileInfo :+1:

jsonsivar (commented):
This looks good so far. It looks like the CI test broke so once we fix that and maybe do some testing like you said, we should be good to merge this.

core/space/domain/domain.go
jsonsivar: No that's fine you can go ahead and remove it. But we should update sharing.go `GetReceivedFiles` to move members into the file info.  I searched across the repo and it seems like that is the only reference to it :+1:

jsonsivar (approved):
LGTM :+1: once we remove the extra `Members` field and then maybe test with grpc or swagger UI the nwe can go ahead and merge this.

dmerrill6 (approved):
LGTM

PR #174, opened 2020-09-11
Hotfix/notifs: don't crash on unknown types and fix subscriptions

jsonsivar:
### Change Log
* Add missing channel creation step in init
* Move watch function into a go routine
* Handle email parsing error for unsupported message types and added UNKNOWN message type


dmerrill6 (approved):
LGTM

PR #175, opened 2020-09-11
Remove CGO_ENABLED env

dmerrill6:
# Change log

- Removed the `CGO_ENABLED=0` env in goreleaser, which was breaking up keychain on runtime. Weird thing is that if I do `CGO_ENABLED=1` it crashes on compilation (as goreleaser does not support CGO). So not sure why removing the env is both fixing the keychain issue and not crashing goreleaser. But anyways, tested a release (v0.0.16-dev) and it completes successfully and fixes the keychain segfault. Tested common daemon operations and they were all working.

[ch18315]

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -8,8 +8,8 @@ before:
 project_name: space
 builds:
   - id: space
-    env:
-      - CGO_ENABLED=0
+    # env:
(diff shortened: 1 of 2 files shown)


jsonsivar (approved):
LGTM

PR #176, opened 2020-09-11
Fix list files decryption

dmerrill6:
# Change log

- Fix decryption of shared files which was missing decrypting the root item when doing `ListPath` of the buckets client.

- Fix a bug I noticed where the textile client never got initialized because ListBuckets was erroring on the mirror bucket.

[ch18274]


perfectmak (approved):
Good catch. LGTM

jsonsivar (approved):

maurycy (approved):
LGTM. 🦸🏻‍♀️

PR #177, opened 2020-09-11
fs: need bucketSlug and bucketPath

maurycy:
The pull request fixes the following bug where updating a file returns the "could not find a bucket with key" error.


dmerrill6 (approved):
LGTM, but please add a PR description and if there's a card to track this bug, the clubhouse tag 🙏

jsonsivar (approved):
LGTM

PR #178, opened 2020-09-12
Fix: panic on Rejecting Invitation

perfectmak:
**CHANGELOG**
- When an invitation is rejected the `createReceivedFiles()` function is called with an empty array of encryption keys and the function assumes all shared items each have a key. This fixes it by conditionally passing in an empty key when not accepted.

- It also adds an invariant check to ensure all shared items have an encryption key present.

### DIFF
--- a/core/textile/sharing.go
+++ b/core/textile/sharing.go
@@ -66,7 +66,7 @@ func (tc *textileClient) AcceptSharedFilesInvitation(
 		return domain.Invitation{}, errInvitationNotPending
 	}
 
-	err := tc.createReceivedFiles(ctx, invitation.InvitationID, true, invitation.ItemPaths, invitation.Keys)
+	err := tc.createReceivedFiles(ctx, invitation, true)
 	if err != nil {
 		return domain.Invitation{}, err
 	}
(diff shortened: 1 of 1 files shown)


core/textile/sharing.go
```diff
@@ -98,23 +98,29 @@ func (tc *textileClient) RejectSharedFilesInvitation(
 
 func (tc *textileClient) createReceivedFiles(
 	ctx context.Context,
-	invitationId string,
+	invitation domain.Invitation,
 	accepted bool,
-	paths []domain.FullPath,
-	keys [][]byte,
 ) error {
+	if len(invitation.ItemPaths) != len(invitation.Keys) {
```
maurycy, line 104: @perfectmak 

My comment would be that it might make sense to create a separate `struct` for sharing the keys and the items _together_, instead of matching them as arguments.

Then, instead of having a special type of empty `encryptionKey` it would be possible to mark the element as rejected.

maurycy (approved):
@perfectmak 

LGTM. 🐎

core/textile/sharing.go
jsonsivar: We could add the key as part of the `FullPath`, but that also means we have to find all the spots that's using this field and update it.  We could make a card for this and add it to upcoming sprints as well since it's working and doesn't affect functionality.

perfectmak: Yeah, I was thinking the same too. It would take more work to refactor for now, so I'll create a card to track it.

PR #180, opened 2020-09-13
Hotfix/backup: show backed up status in listdirs queries

jsonsivar:
Fixes https://app.clubhouse.io/terminalsystems/story/18278/bug-wire-up-backed-up-field-in-list-items


grpc/handlers.go
```diff
@@ -98,6 +98,11 @@ func (srv *grpcServer) ListDirectory(
 			})
 		}
 
+		var backupCount = 0
+		if e.BackedUp {
+			backupCount = 1
```
maurycy, line 103: @jsonsivar 

Is that a correct counter?

core/space/domain/domain.go
```diff
@@ -26,6 +26,7 @@ type ThreadInfo struct {
 type FileInfo struct {
 	DirEntry
 	IpfsHash string
+	BackedUp bool
```
maurycy, line 29: @jsonsivar 

What do you think about renaming this to `Backup` so we have the same convention as in the `Bucket` `struct`?

maurycy (approved):
LGTM. 🍣

grpc/handlers.go
jsonsivar: Yeah for now it is, we are storing it as a bool internally but didn't change that yet because we can do that with the broader piece of work that's upcoming related to that (online/offline state management and multi-hub support).

core/space/services/services_fs.go
```diff
@@ -530,7 +542,12 @@ func (s *Space) addFile(ctx context.Context, sourcePath string, targetPath strin
 
 	_, fileName := filepath.Split(sourcePath)
 
-	targetPathBucket := targetPath + "/" + fileName
+	var targetPathBucket string
+	if targetPath == "" {
```
perfectmak, line 546: what of situation where a user specifies targetPath as `"/"` and not empty string. The `b.UploadFile()` already handles this edge case but how does it affect this if statement here? Maybe you should also check for paths only having a path separator.

perfectmak (approved):
LGTM! I left a comment.

core/textile/model/mirror_file.go
```diff
@@ -59,6 +60,43 @@ func (m *model) CreateMirrorBucket(ctx context.Context, bucketSlug string, mirro
...
+		return nil, err
+	}
+
+	if rawMirrorFiles == nil {
+		return nil, nil
+	}
+
+	mirror_files := rawMirrorFiles.([]*MirrorFileSchema)
+	if len(mirror_files) == 0 {
+		return nil, nil
```
dmerrill6, line 89: Maybe these should return a "path not found" error.

dmerrill6 (approved):
LGTM

core/textile/model/mirror_file.go
jsonsivar: The thing is if this returns an error instead, it will break the list dirs call and I was thinking it should still return what it can with list dirs so the files at least show up even if the mirror info wasn't able to be fetched.  Also the similar function `FindMirrorFileByPathAndBucketSlug` does the same so it stays consistent with that. In that case we check that before creating and if its empty we go ahead and create a mirror file (if we return error there it would error our add files).  I guess the other option is to return `pathNotFound` in both cases but handle the error in the calling code but figured it's okay for a `find` call to return 0 results but not be an error state similar to the underlying threads Find call.

core/space/services/services_fs.go
jsonsivar: Fixed in [8ccc69f](https://github.com/FleekHQ/space-daemon/pull/180/commits/8ccc69f2f66636b1462fbe89caf17f0d6b2252ff).

core/textile/model/mirror_file.go
maurycy: @dmerrill6 @jsonsivar 

Maybe we can revisit these later and add `skipOnError` or similar flag?

dmerrill6: @maurycy 
Don't really like how optional args are handled in Go so feels a bit overkill to change the method return given that flag. In that case we could do some nomenclature agreement like `FetchX` errors if X is not found and `FindX` returns nil if not found. 

@jsonsivar 
I was thinking returning `nil, nil` with a map in the response could lead to errors if not handled on the caller side (e.g. if the caller does `if err == nil then doSomething(map[key])` would panic). But yea it's the same issue for the other Find method so I think we can revisit later and update the callers to do something like `if err.name == "ErrNotFound" then continue`.

jsonsivar: That's a good point. I like aligning on that nomenclature as well. Will create a card to clean that up and put it in our beta board.

PR #182, opened 2020-09-14
Hotfix/sharing: show members on received files

jsonsivar:
Fixes https://app.clubhouse.io/terminalsystems/story/18271/bug-get-members-on-received-files

### Change Log
* Fix on textile side: https://github.com/textileio/textile/pull/333 (note this is already updated in our dev hub so it was tested)
* Save current user as owner when adding a file (otherwise sender does not show in recipients members list on that file)


core/space/services/services_fs.go
```diff
@@ -548,6 +548,15 @@ func (s *Space) addFile(ctx context.Context, sourcePath string, targetPath strin
 			return domain.AddItemResult{}, err
 		}
 
+		//add current user as admin on the remote file
+		err = s.tc.AddCurrenUserAsFileOwner(ctx, b, targetPathBucket)
+		if err != nil {
+			log.Error("Error adding current user as file owner", err)
```
maurycy, line 554: @jsonsivar 

It looks like a good candidate for _optional_ error tracking (eg: [Sentry](https://docs.sentry.io/platforms/go/)) in the future.

maurycy (approved):
LGTM. 🔌

core/space/services/services_fs.go
```diff
@@ -548,6 +548,15 @@ func (s *Space) addFile(ctx context.Context, sourcePath string, targetPath strin
 			return domain.AddItemResult{}, err
 		}
 
+		//add current user as admin on the remote file
+		err = s.tc.AddCurrenUserAsFileOwner(ctx, b, targetPathBucket)
```
dmerrill6, line 552: Maybe this can be an internal call of the Textile Client that gets called inside `UploadFileToHub` to avoid adding more methods to the public interface of the Client.

dmerrill6 (approved):
LGTM, left a minor comment but maybe it's better the way it is now.

core/space/services/services_fs.go
jsonsivar: Good call.  Fixed in [a8ad2ff](https://github.com/FleekHQ/space-daemon/pull/182/commits/a8ad2fffb5d31c798ecc794ffa82718ab9852542).

PR #186, opened 2020-09-15
Make textile client dependent methods more robust

dmerrill6:
# Change log

- Make Space Service wait for Textile Client init/hub connection before executing methods such as list dir, get notifs, etc. In case of a timeout or 3 consecutive failed connections, return an error.

## Notes

This flow was discussed with @giankotarola yesterday and we agreed it would allow the frontend to better handle wait times, as the FE can display the loading states of each component while textile is not ready (e.g. show skeletons in notifs or file list), and only fail execution if there's a deeper error such as no internet connection.

Also this fixes the issue where if you log out and log in into the same account, it didn't display the existing files. Now that the client returns the response after textile is initialized, the FE is receiving the files and displaying them correctly.

@giankotarola: I tested this version using the electron app and it's working fine, but I'm not able to see the loading components (for example something to display that the list of files is loading). Maybe that needs to be implemented?

[ch18397]


giankotarola:
@dmerrill6 cool! And yea, it was implemented but had to be reverted since it had a bug, but def will be fixed soon.

core/space/services/services.go
```diff
@@ -68,3 +70,44 @@ func NewSpace(
...
+// Waits for textile client to be initialized before returning.
+func (s *Space) waitForTextileInit(ctx context.Context) error {
+	if s.tc.IsInitialized() {
+		return nil
+	}
+
+	select {
+	case <-time.After(textileClientInitTimeout):
+		return errors.New("textile client not initialized in expected time")
+	case <-s.tc.WaitForInitialized():
```
maurycy, line 86: @dmerrill6 

Should we `s.tc.isInitialized = true` here?

Then, the `tc.onInitialized` chan would encapsulate the internal state.

maurycy (approved):
LGTM. 💠

jsonsivar (approved):

perfectmak (approved):

core/space/services/services.go
dmerrill6: We are already doing that internally inside the `initialize` method of the Textile client, unless I'm misunderstanding what you are trying to say

maurycy: @dmerrill6 

You are right.

PR #187, opened 2020-09-16
Hotfix/sharing: bugs related to opening received file

jsonsivar:
### Change Log
* Before it was failing with `user does not own thread` because we were calling `list` buckets but the receiver doesn't have permissions to list all buckets since they only got shared a file (instead of listing all buckets, we now just fetch the path that was shared, and use it's `GetRoot` func to get the bucket root)
* A leftover log that was accessing an array item that was outside the index range - removed that altogether since it was just used for logging
* Added example of sharing and accessing file that helped isolate the issue


dmerrill6 (approved):
Nice catch, LGTM

PR #189, opened 2020-09-17
Hotfix/ch18277: show accept/reject on invitations

jsonsivar:
### Change Log
* Get received files by ID helper
* Make message parser handle multiple so we can keep message specific logic in there
* Add fetch of statuses for messages
* Also a couple bugs that was in the way of this working:
* Bug: notification ID was not being saved in received files
* Bug: GetThreadContext was not creating new token so if transitioning from hub to local it failed (hotfixed this for now but maybe in a future fix we can tie this back to the cached token and switch based on the passed in `threadsClient` host or something)

### DIFF
--- a/core/textile/utils/utils.go
+++ b/core/textile/utils/utils.go
@@ -11,6 +11,7 @@ import (
 	"github.com/FleekHQ/space-daemon/core/keychain"
 	"github.com/FleekHQ/space-daemon/core/textile/hub"
 	crypto "github.com/libp2p/go-libp2p-crypto"
+	tc "github.com/textileio/go-threads/api/client"
 	"github.com/textileio/go-threads/core/thread"
 	"github.com/textileio/textile/api/common"
 	"golang.org/x/crypto/pbkdf2"
@@ -80,7 +81,7 @@ func getThreadName(userPubKey []byte, bucketSlug string) string {
 }
 
 // Readies a context to access a thread given its name and dbid
-func GetThreadContext(parentCtx context.Context, threadName string, dbID thread.ID, hub bool, kc keychain.Keychain, hubAuth hub.HubAuth) (context.Context, error) {
+func GetThreadContext(parentCtx context.Context, threadName string, dbID thread.ID, hub bool, kc keychain.Keychain, hubAuth hub.HubAuth, threadsClient *tc.Client) (context.Context, error) {
 	var err error
 	ctx := parentCtx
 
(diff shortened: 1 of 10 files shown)


maurycy (approved):

core/textile/utils/utils.go
```diff
@@ -94,7 +95,8 @@ func GetThreadContext(parentCtx context.Context, threadName string, dbID thread.
 	}
 
 	var publicKey crypto.PubKey
-	if publicKey, err = kc.GetStoredPublicKey(); err != nil {
+	var privKey crypto.PrivKey
+	if privKey, publicKey, err = kc.GetStoredKeyPairInLibP2PFormat(); err != nil {
```
dmerrill6, line 99: Is this really needed? Because this means it will ask for the keychain pass again

dmerrill6 (approved):
LGTM, left a comment tho, but maybe we can refactor later by storing the priv key to memory after getting it for the first time.

core/textile/utils/utils.go
jsonsivar: Yeah it fails without this.  Because the token in the context could be from a previous hub ctx e.g. this fails with invalid thread token if we don't create a new token.  And to create the new token it expects a privkey. It's related to the last bug in the change log.

dmerrill6: Makes sense, but this will become really annoying on OSX 😅 
I think we need to include a cache in `keychain.go` so that it pulls/stores the private key from memory.

I can take a stab at this later in the afternoon if you agree with that.

jsonsivar: Fair enough, I'll keep this PR open until we add that then @dmerrill6. I can also take a stab at it on this branch.  If so I'm thinking of adding the privkey as a field on the `kc` instance that's gets set on startup. And a helper function to return that value. Does anything else need to be done or you had something else in mind?

dmerrill6: That sounds good. I think the current getter (`GetStoredKeyPairInLibP2PFormat`) can look if the value is set and return it, or do the current flow otherwise. Also `GenerateKeyPair, GenerateKeyFromMnemonic, GenerateKeyPairWithForce, ImportExistingKeyPair, DeleteKeypair` need to clear this cache.

jsonsivar: @dmerrill6 please review the following commit to address this https://github.com/FleekHQ/space-daemon/pull/189/commits/caa6f8bee2c20a29d2c88cfe795b28c0cd0a34f5.

* `GetStoredKeyPairInLibP2PFormat` like you said checks for the cache and returns that, if not found sets it
* Added the save to cache in `generateAndStoreKeyPair` so it handles all of `GenerateKeyPair`, `GenerateKeyPairWithForce` and `GenerateKeyFromMnemonic`
* Update/remove cache in `ImportExistingKeyPair`, `DeleteKeypair`
* Assuming `GetStoredPublicKey` was working without pw prompt so didn't touch that

dmerrill6: Yea GetStoredPublicKey shouldn't change. The changes LGTM 👍

PR #190, opened 2020-09-18
fix: missing marks in the reflection tags

maurycy:


perfectmak (approved):
Oh wow! How did we miss this? Thanks.

We should set up linters for things like this as part of our CI process.

PR #191, opened 2020-09-21
backup: additional toggle logic

maurycy:
The pull request adds additional toggle logic for backing up files, and prepares for replication of threads.


core/textile/backup.go
```diff
@@ -0,0 +1,266 @@
...
+
+	if _, err = tc.setMirrorFileBackup(ctx, path, bucket.Slug()); err != nil {
+		log.Error(fmt.Sprintf("error setting mirror file as backup (path=%+v b.Slug=%+v)", path, bucket.Slug()), err)
+		return err
+	}
+
+	return nil
+}
+
+// backup a single file with an io.Reader provided
```
dmerrill6, line 31: This docstring is off

core/textile/backup.go
```diff
@@ -0,0 +1,266 @@
...
+				errc <- err
+				return
+			}
+
+			count += 1
+		}(item.Path)
+	}
+
+	wg.Wait()
+
```
dmerrill6, line 104: Why not `<- errc` here? Or is this supposed to be kind of a background process?

core/textile/backup.go
```diff
@@ -0,0 +1,266 @@
...
+			count += n
+			continue
+		}
+
+		// parallelize the backups
+		go func(path string) {
+			defer wg.Done()
+
+			if err = tc.backupFile(ctx, bucket, path); err != nil {
+				errc <- err
```
dmerrill6, line 95: What happens if there's no one listening at `errc`? I've seen Go get stuck in those cases, that's why I've had to do stuff like 
```go
select {
  case errc <- err:
    log.Debug("Msg sent")
  default:
    // Do nothing
}
```

Maybe since this is within a go subroutine it doesn't matter but I'd make sure it doesn't hang when for example `Ctrl+C`ing the daemon.

core/textile/backup.go
```diff
@@ -0,0 +1,266 @@
...
+func (tc *textileClient) unbackupBucketFiles(ctx context.Context, bucket Bucket, path string) (int, error) {
+	var wg sync.WaitGroup
+	var count int
+
+	dir, err := bucket.ListDirectory(ctx, "")
+	if err != nil {
+		return 0, err
+	}
+
+	wg.Add(len(dir.Item.Items))
```
dmerrill6, line 202: Not sure if this is right. We are just adding one item to the working group (since this is not inside the for) ,so this is just going to continue past `wg.Wait()` after the first item does the `wg.Done()`.

core/textile/bucket_factory.go
```diff
@@ -119,12 +118,19 @@ func (tc *textileClient) getOrCreateBucketContext(ctx context.Context, bucketSlu
...
 
+	managedKey, err := tc.kc.GetManagedThreadKey()
+	if err != nil {
+		log.Error("error getting managed thread key", err)
+		return nil, nil, err
+	}
+
 	log.Debug("getOrCreateBucketContext: Creating Thread DB for bucket " + bucketSlug + " at db " + dbID.String())
-	if err := tc.threads.NewDB(ctx, dbID); err != nil {
+	if err := tc.threads.NewDB(ctx, dbID, db.WithNewManagedThreadKey(managedKey)); err != nil {
```
dmerrill6, line 128: What's this for?

core/textile/sharing.go
```diff
@@ -273,3 +273,27 @@ func (tc *textileClient) GetPathAccessRoles(ctx context.Context, b Bucket, path
...
 }
+
+// return true if file was shared
+// XXX: export this func?
+func (tc *textileClient) isSharedFile(ctx context.Context, bucket Bucket, path string) bool {
+	sbc := NewSecureBucketsClient(tc.hb, bucket.Slug())
+
+	roles, err := sbc.PullPathAccessRoles(ctx, bucket.Key(), path)
+	if err != nil {
+		// XXX: returning empty members list until we fix it on textile side
```
dmerrill6, line 284: This comment is off

maurycy: @dmerrill6 Thank you for spotting this!

core/textile/bucket_factory.go
maurycy: @dmerrill6 

My intuition was to manage keys for every thread that's not public. What do you think, @jsonsivar ?

core/textile/backup.go
maurycy: @dmerrill6 Thank you for spotting this!

core/textile/backup.go
perfectmak: I don't think select is necessary in this case @dmerrill6, since in this case we are sending `err` through the `errc` and not waiting for a value from `errc`.Also, it wouldn't block because of the same reason.

core/textile/backup.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// backup a single file with a path provided
+func (tc *textileClient) backupFile(ctx context.Context, bucket Bucket, path string) error {
+
+	errc := make(chan error, 1)
+	pipeReader, pipeWriter := io.Pipe()
+
+	// go routine for piping
+	go func() {
```
perfectmak, line 38: Why run this as a goroutine? It looks like you could has well just run it inline with the current execution stack and return any failed error and therefore won't be need for an error channel?

core/textile/backup.go
```diff
@@ -0,0 +1,266 @@
...
+
+	return nil
+}
+
+// backup all files in a bucket
+func (tc *textileClient) backupBucketFiles(ctx context.Context, bucket Bucket, path string) (int, error) {
+	var wg sync.WaitGroup
+	var count int
+
+	// XXX: we ignore errc (no atomicity at all) for now but we should return
```
perfectmak, line 66: Maybe we should think through a more atomic/consistent way of doing backups, for example if something goes wrong during a backup, do we plan to start all over again or we should actually track what failed and only backup those?

core/textile/backup.go
```diff
@@ -0,0 +1,266 @@
...
+	dbIds = append(dbIds, publicShareId)
+
+	return dbIds, nil
+}
+
+// replicate meta, key and shared with me threads
+func (tc *textileClient) replicateThreadsToHub(ctx context.Context, bucket Bucket) (count int, err error) {
+	replicatedDbIds := make([]thread.ID, 0)
+
+	// poor man's transactionality
```
perfectmak, line 133: I know threads have transactions capabilities backed in, maybe you can look into that so it would be easier to commit changes or rollback if there is an error.

perfectmak (commented):
I see you still have some work to be done on this PR, but I left some comments on my observation. When it is ready, I can re-review

core/textile/backup.go
maurycy: @perfectmak 

This is a `io.Pipe` pattern. `bucket.GetFile()` writes to the `pipeWriter` and, at the same time, it is being read by `tc.BackupFileWithReader()`, so we don't need any temporary files. Please take a look at the [example](https://golang.org/pkg/io/#Pipe).

perfectmak: Yes, I know of the pattern, but it still looks like it would still be executed sequentially expecially with the `if err := <-errc` used immediately after the go func that is supposed to write to the `pipeWriter`. And then after the err check you pass `pipeReader` into `tc.BackupFileWithReader()`.

If you have sometime, you should try testing it with the go routine and without and see if there is any difference.

core/textile/backup.go
maurycy: @perfectmak 
@dmerrill6 

It is a buffered channel (`make(chan, 1)`), so it blocks only when the buffer is full (ie: in this case if `err` is sent but not received.)

maurycy: @dmerrill6 @perfectmak 

I've added the `select` to avoid any confusion in https://github.com/FleekHQ/space-daemon/pull/191/files/593541272f2e1edb98679e8020351e5a6ecb5232#diff-b580f3c0539be5a72155b66b76313a90R103 (Please ignore my previous comment about the buffered channel.)

core/textile/backup.go
maurycy: @perfectmak 

Great catch!

Unfortunately, we're calling https://github.com/textileio/go-threads/blob/8355a2e43aa1abc8b919851b3714abc0ce1d41c9/net/api/client/client.go#L200 under the hood so https://github.com/textileio/go-threads does not seem to offer transactions here, unless I'm misreading the code.

core/textile/backup.go
maurycy: @dmerrill6 

Thank you for spotting this! This code was incorrect anyway because there is `.textile` and `.textileseed`. :-)

ref https://github.com/FleekHQ/space-daemon/pull/191/files/593541272f2e1edb98679e8020351e5a6ecb5232#diff-b580f3c0539be5a72155b66b76313a90R232

core/textile/backup.go
maurycy: @perfectmak 

I don't know. I don't want to delay this issue any longer but I agree we should decide on atomicity. What do you think, @jsonsivar @dmerrill6 ?

core/textile/backup.go
maurycy: @dmerrill6 

Thank you for the comment. Mostly because I don't want to overload `errc` (it is for future error handling, not for coordination.)

core/textile/backup.go
maurycy: @perfectmak 

Great find. Thank you. I've tweaked this code so now it's truly parallel: https://github.com/FleekHQ/space-daemon/pull/191/files/593541272f2e1edb98679e8020351e5a6ecb5232#diff-b580f3c0539be5a72155b66b76313a90R53

core/textile/bucket_factory.go
dmerrill6: Before approving, need to fully understand what's this for since you are using the public key for this value, but given it's public it can be generated by anyone which I'm not sure if it brings issues like collisions. If you check `textile/utils` in the method `NewDeterministicThreadID` we derive the thread id using a KDF from the private key, which would solve this issue if it is.

jsonsivar: Yeah the thread key should be the private key or something derived from it but not a public key. @maurycy saw the fix commit for this but I think that should be using the private key instead of the public key.

core/textile/backup.go
jsonsivar: It would be nicer to not have to start all over again. So maybe there is a backup state (not started, in progress, up-to-date) that we can set and sync based off.

core/textile/bucket_factory.go
maurycy: @jsonsivar 

We will be syncing about this later today, because it turned out that `NewDeterministicThreadID` is not using the private key as well. :-)

core/textile/backup.go
maurycy: @jsonsivar 

We're keeping the successful state via `setMirrorFileBackup` and `unsetMirrorFileBackup`. I've added `isMirrorBackupFile` in `backupBucketFiles` in 7cb351cedbc82052b2cdbba91c391b277c9b2160 to avoid starting all over again (waiting for the hub to recover so I can test it), that should be good for now.

core/textile/bucket_factory.go
jsonsivar: I think it might be okay that the thread ID creation is using the pubkey since that is just the thread ID.  Even if someone tries to generate it using the pubkey and then tries accessing, it wont give them access because it was created with the private key context.  Here we are creating a thread key (not thread ID) hence the difference and need to definitely use the private key.  Just clarifying the difference, but we can still make the thread ID also use the private key as well for consistency but it is 2 different things we are doing here.

maurycy: @jsonsivar 

What do you think? https://github.com/FleekHQ/space-daemon/pull/191/commits/2bf0da6c2f418e5412ed4f632666eae22c6ade2e#diff-a509c084f1fb203e2d90b6bd7e71b528R377

jsonsivar: Nice looks good.

core/keychain/keychain.go
```diff
@@ -366,3 +370,24 @@ func (kc *keychain) retrieveKeyPair() (privKey []byte, mnemonic string, err erro
...
+
+func (kc *keychain) GetManagedThreadKey(threadKeyName string) (thread.Key, error) {
+	size := 32
+
+	privBytes, _, err := kc.retrieveKeyPair()
+	if err != nil {
+		return thread.Key{}, err
+	}
+
+	num := pbkdf2.Key(privBytes, []byte("threadID"+threadKeyName), 256, size, sha512.New)
```
dmerrill6, line 382: Just nitpick but this should be `"ThreadKey" + threadKeyName`

dmerrill6 (approved):
Left a small correction but other than that LGTM

core/keychain/keychain.go
maurycy: @dmerrill6 😸 

https://github.com/FleekHQ/space-daemon/pull/191/commits/5a0a95a44edc6a31d21851f63ecd2ed0a68c5f5c#diff-a509c084f1fb203e2d90b6bd7e71b528R382

core/keychain/keychain.go
```diff
@@ -366,3 +370,24 @@ func (kc *keychain) retrieveKeyPair() (privKey []byte, mnemonic string, err erro
 
 	return privKey, mnemonic, nil
 }
+
+func (kc *keychain) GetManagedThreadKey(threadKeyName string) (thread.Key, error) {
+	size := 32
+
+	privBytes, _, err := kc.retrieveKeyPair()
```
jsonsivar, line 377: Seeing @dmerrill6 's other PR reminded me we should use `GetStoredKeyPairInLibP2PFormat` here as well so it checks the cache first under the hood.

maurycy: @jsonsivar 

Thank you! That's the reason for the slowness I noticed then.

f0b4486a678384dc96c4ce8af3d70f8ebb3aa927

jsonsivar:
LGTM.  Re: backup-in-progress states, it sounded like we need a bucket level status field (not started, in progress, complete, error) and then check those statuses during toggles so we don't kick off multiple jobs. I think this was related to @perfectmak and he had requsted a re-review so if we close the loop on that, we can merge this and handle that in the next card.

perfectmak (approved):

PR #193, opened 2020-09-21
Allow generating public link for shared files

dmerrill6:
# Change log

- Add `dbID` param to `GenerateFilesSharingLink` to enable generating public links for shared files

[ch18164]


perfectmak (approved):
LGTM

PR #196, opened 2020-09-25
Offline support – Alpha

dmerrill6:
# Change log

- Add background process that syncs local updates to remote (hub)

- This process manages a queue of tasks for each of the pending syncing tasks.

- Refactor mirror bucket logic to be fully handled by the synchronizer process.

There's still stuff missing before considering this PR complete, but wanted to push anyways in case this helps upcoming tasks to follow the new approach.

## TODO

- Tests

- Migrate public bucket functions to synchronizer

- Currently task failures are retried after 30 seconds. Need to add some kind of exponential backoff.

- Update proto schema to return a "Pending upload" state so that the FE can show a hint when a file is available locally but still being backed up.

- Send notifications through grpc whenever there's a state change in the backed up state.


core/textile/bucket/bucket.go
```diff
@@ -74,15 +115,23 @@ func (b *Bucket) GetData() BucketData {
...
+	_, threadID, err := b.GetContext(ctx)
 	if err != nil {
 		return nil, err
 	}
 
 	return threadID, nil
 }
+
+func (b *Bucket) AttachNotifier(n Notifier) {
+	b.notifier = n
```
maurycy, line 136: @dmerrill6 

Should it be possible to attach more than one notifier, eg: one-to-many pins?

core/textile/sync/queue.go
```diff
@@ -0,0 +1,132 @@
...
+	if s.queueHashMap[task.ID] != nil {
+		return true
+	}
+
+	return false
+}
+
+func (s *synchronizer) printQueueStats(queue *list.List) {
+	queueName := "buckets"
+	if queue == s.filePinningQueue {
```
maurycy, line 112: @dmerrill6 

What do you think about `(s *synchronizer) String()`?

core/textile/sync/task.go
```diff
@@ -0,0 +1,53 @@
...
+	Args           []string  `json:"args"`
+	Parallelizable bool      `json:"parallelizable"`
+
+	// Set to -1 for infinite retries
+	MaxRetries int `json:"maxRetries"`
+	Retries    int `json:"retries"`
+}
+
+func newTask(t taskType, args []string) *Task {
+	id := uuid.UUID1().String()
```
maurycy, line 42: @dmerrill6 

What do you think about having the `id` that is a hash of the task `args`, so there is deduplication built-in?

core/textile/sync/queue.go
```diff
@@ -0,0 +1,132 @@
...
+const QueueStoreKey = "TextileSyncTaskQueue"
+
+type marshalledQueue struct {
+	queueAsSlice     []*Task          `json:"queueAsSlice"`
+	fileQueueAsSlice []*Task          `json:"fileQueueAsSlice"`
+	hashMap          map[string]*Task `json:"queueHashMap"`
+}
+
+func (s *synchronizer) enqueueTask(task *Task, queue *list.List) {
+	queue.PushBack(task)
```
maurycy, line 20: @dmerrill6 

Do you think we can simplify `container/list` to just a slice?

core/textile/sync/queue.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for currEl != nil {
+		fileQueueAsSlice = append(fileQueueAsSlice, currEl.Value.(*Task))
+		currEl = currEl.Next()
+	}
+
+	objToMarshal := &marshalledQueue{
+		queueAsSlice:     queueAsSlice,
+		fileQueueAsSlice: fileQueueAsSlice,
+		hashMap:          s.queueHashMap,
```
maurycy, line 56: @dmerrill6 

Does it duplicate the same `Task` as in the slices?

What is the plan for `hashMap` except for a (unused) `isTaskEnqueued` check?

core/textile/sync/synchronizer.go
```diff
@@ -0,0 +1,304 @@
...
+			}
+		}
+
+		if task.Parallelizable && parallelTaskCount < maxParallelTasks {
+			parallelTaskCount++
+
+			go func() {
+				err := s.executeTask(ctx, task)
+				handleExecResult(err)
+				parallelTaskCount--
```
maurycy, line 267: @dmerrill6 

Do you think this is a race condition? I think that [atomic](https://golang.org/pkg/sync/atomic/) would be helpful here.

core/textile/bucket/bucket.go
dmerrill6: Don't see a use for that right now, given that if you wanted to do multiple pins you could simply write a custom notifier that does that. We can modify it in the future anyways in case we want to have multiple notifiers.

core/textile/sync/queue.go
dmerrill6: Thought about it but ended up going for this instead because it's not really the synchronizer state but one of the queues state.

core/textile/sync/task.go
dmerrill6: Good idea, added that in 79c4212

core/textile/sync/queue.go
dmerrill6: I chose using a linked list because it's perfect to represent a queue, as queue/dequeue operations are O(1) and memory efficient. Not sure really how Go handles slices internally when doing stuff like `slice = slice[:-1]` but I rather have control.

core/textile/sync/queue.go
dmerrill6: > Does it duplicate the same Task as in the slices?
Not sure what you mean by this

What is the plan for hashMap except for a (unused) isTaskEnqueued check?
I did it initially in case I wanted to access an element in the queue directly, so that I could get access in O(1) instead of finding it in the linked list, but ended up not using it. However after implementing your suggestion about task IDs, I'm now using it when enqueuing a new task. First it will see if the task exists in the queue and avoid adding it if it's already there.

core/textile/sync/synchronizer.go
dmerrill6: There shouldn't be any race condition issues as `sync` func is behind a lock. However after this comment I noticed this var was being shared by both queues so I fixed that in 10b65a3

core/textile/sync/queue.go
dmerrill6: Stopped storing the hash map in f4d3cc6

dmerrill6:
Added a few tests and cleaned out some bugs. Moved from draft to ready.

jsonsivar (approved):
LGTM. Nice work!

core/textile/bucket/bucket_file.go
```diff
@@ -44,19 +44,29 @@ func (b *Bucket) FileExists(ctx context.Context, pth string) (bool, error) {
...
 	}
-	return b.bucketsClient.PushPath(ctx, b.Key(), path, reader)
+
+	result, root, err = b.bucketsClient.PushPath(ctx, b.Key(), path, reader)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if b.notifier != nil {
+		b.notifier.OnUploadFile(b.Slug(), path, result, root)
```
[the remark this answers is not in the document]
dmerrill6: Not sure I understand fully but `bucket` doesn't have access to `tc` so where would that `tc.sync.Notify...` call be?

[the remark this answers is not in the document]
jsonsivar: I guess that should have said `b.notifier.sync.Notify...` but maybe the caller doesn't have permissions to do that. I was just curious why add file went through this but add bucket and the others didn't.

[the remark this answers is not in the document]
dmerrill6: I see. I tried that inside the bucket by passing the sync directly but that ended up throwing a dependency cycle error, so I had to do it this way through an interface

PR #198, opened 2020-09-28
Sync alpha with master

jsonsivar:
This brings some commits history that happened on `master` but weren't included in the alpha branch (this is a precursor to merging `alpha-release-dev` to `master`).


maurycy:
@jsonsivar 

What do you think about using `Squash and merge` in this case, so the commit log is cleaner?

jsonsivar:
> @jsonsivar
> 
> What do you think about using the `Squash and merge` in this case, so the commit log is cleaner?

Sounds good to me.

dmerrill6 (approved):
LGTM

PR #202, opened 2020-10-02
Add backup progress status and notification

dmerrill6:
# Change log

- Add "backup in progress" state to mirror files

- Notify client when it starts and completes


maurycy (approved):
LGTM. 🚀

PR #204, opened 2020-10-07
parametrize Service.ListDir(s)()

maurycy:
This enables frontend to control loading of the listing and optimize the performance, without breaking backwards compatibility.


dmerrill6 (commented):
I think the `omiSubfolderContent` is not necessary, since we already have the `ListDirectories` and `ListDirectory` grpc (first one includes subfolders, second one not).

Maybe another small improvement would be to use optional parameters for these flags, so you would call the service like `s.ListDir(..., s.ListDirWithMembers())`

maurycy:
@dmerrill6 

Thank you! Removed `{omit,list}SubfolderContent`. We might want to merge `ListDirectory()` and `ListDirectories()` eventually but it's not important right now.

dmerrill6 (approved):
LGTM

PR #207, opened 2020-10-09
restore files

maurycy:
The pull request adds support for restoring files that were backed up. :-)

[ch17924]
[ch17935]
[ch17036]


core/events/events.go
```diff
@@ -15,6 +15,10 @@ const (
 	FileBackupInProgress FileEventType = "FileBackupInProgress"
 	FileBackupReady      FileEventType = "FileBackupReady"
 
+	FileRestoreInProgress FileEventType = "FileRestoreInProgress"
+	FileRestored          FileEventType = "FileRestored"
+	FileRestoring         FileEventType = "FileRestoreInProgress"
```
dmerrill6, line 20: The string value is duplicated

core/textile/bucket/bucket.go
```diff
@@ -47,6 +47,11 @@ type BucketInterface interface {
 		path string,
 		reader io.Reader,
 	) (result path.Resolved, root path.Path, err error)
+	DownloadFile(
```
dmerrill6, line 50: Why is this different from GetFile?

core/textile/bucket/bucket_file.go
```diff
@@ -61,6 +61,27 @@ func (b *Bucket) UploadFile(ctx context.Context, path string, reader io.Reader)
 	return result, root, nil
 }
 
+func (b *Bucket) DownloadFile(ctx context.Context, path string, reader io.Reader) (result path.Resolved, root path.Path, err error) {
```
dmerrill6, line 64: DownloadFile to me means bringing a file on the remote to a local file (GetFile), but this is doing the same that `UploadFile` except it is additionally notifying the syncer. Shouldn't we just modify UploadFIle?

core/textile/sync/mirror.go
```diff
@@ -24,7 +24,7 @@ func (s *synchronizer) setMirrorFileBackup(ctx context.Context, path, bucketSlug
 	}
 	if mf != nil {
 		// update
-		mf.Backup = !isInProgress
+		mf.Backup = true
```
dmerrill6, line 27: This will make files that are not yet pinned be marked as pinned

core/textile/sync/pinning.go
```diff
@@ -55,6 +54,51 @@ func (s *synchronizer) uploadFileToRemote(ctx context.Context, bucket, path stri
 	return nil
 }
 
+func (s *synchronizer) downloadFileFromRemote(ctx context.Context, bucketSlug, path string) error {
```
dmerrill6, line 57: Maybe we can refactor `uploadFileToRemote` to use an auxiliary function used by both that method and this one instead of duplicating the function but swapping two vars

core/textile/sync/restore.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	mirrorFiles, err := s.model.FindMirrorFileByPaths(ctx, dirPaths)
+	if err != nil {
+		log.Error("Error fetching mirror files", err)
+		return err
+	}
+
+	for _, m := range mirrorFiles {
+		if err = s.downloadFileFromRemote(ctx, m.BucketSlug, m.Path); err != nil {
```
dmerrill6, line 37: Maybe an option we can do is span a "DownloadFile" task for each file instead of doing it all in one task

grpc/handlers_backup.go
```diff
@@ -6,17 +6,28 @@ import (
...
-	err := srv.sv.ToggleBucketBackup(ctx, bucketName, bucketBackup)
+	err := srv.sv.ToggleBucketBackup(ctx, bucketSlug, bucketBackup)
 	if err != nil {
 		return nil, err
 	}
 
 	return &pb.ToggleBucketBackupResponse{}, nil
 }
 
+func (srv *grpcServer) BucketBackupRestore(ctx context.Context, request *pb.BucketBackupRestoreRequest) (*pb.BucketBackupRestoreResponse, error) {
```
dmerrill6, line 20: I think this should be an internal call when starting up the daemon instead of being triggered by a client. That way we avoid adding a button to start the process when it should be handled in the background. What do you think?

core/events/events.go
maurycy: @dmerrill6 Thank you for spotting this!

core/events/events.go
```diff
@@ -15,6 +15,10 @@ const (
 	FileBackupInProgress FileEventType = "FileBackupInProgress"
 	FileBackupReady      FileEventType = "FileBackupReady"
 
+	FileRestoreInProgress FileEventType = "FileRestoreInProgress"
+	FileRestored          FileEventType = "FileRestored"
+	FileRestoring         FileEventType = "FileRestoring"
```
jsonsivar, line 20: What's the difference between `FileRestoring` and `FileRestoreInProgress`? Maybe we can call the formed `FileRestoreStarted`?

core/textile/sync/restore.go
```diff
@@ -0,0 +1,40 @@
+package sync
+
+import (
+	"context"
+
+	"github.com/FleekHQ/space-daemon/log"
+)
+
+// replicate a local thread on the hub
```
jsonsivar, line 9: This is more so replicated hub data locally right?

core/textile/sync/restore.go
```diff
@@ -0,0 +1,40 @@
...
+		log.Error("Error in ListDirectory", err)
+		return err
+	}
+
+	dirPaths := make([]string, 0)
+	for _, item := range dir.Item.Items {
+		dirPaths = append(dirPaths, item.Path)
+	}
+
+	mirrorFiles, err := s.model.FindMirrorFileByPaths(ctx, dirPaths)
```
jsonsivar, line 29: This will make sure any existing files are updated to the latest version.  But we still need something to browse remote bucket and add any missing files to the local bucket right? Maybe that's the other piece we talked about in standup that needs to be done.

core/textile/sync/synchronizer.go
```diff
@@ -243,6 +257,8 @@ func (s *synchronizer) executeTask(ctx context.Context, t *Task) error {
 		err = s.processBucketBackupOn(ctx, t)
 	case bucketBackupOffTask:
 		err = s.processBucketBackupOff(ctx, t)
+	case bucketRestoreTask:
```
jsonsivar, line 260: Do we need one for the file task as well?

core/textile/sync/task-executors.go
```diff
@@ -172,3 +182,48 @@ func (s *synchronizer) processBucketBackupOff(ctx context.Context, task *Task) e
...
+	}
+
+	mirrorBucket, err := s.getMirrorBucket(ctx, bucket)
+	if err != nil {
+		return err
+	}
+
+	// TODO: use timestamp or CID for check
+
+	if err = s.uploadFileToRemoteBucket(ctx, mirrorBucket, localBucket, path); err != nil {
```
jsonsivar, line 220: This should upload the file locally right? Maybe just a function naming thing.

core/textile/sync/restore.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+
+	"github.com/FleekHQ/space-daemon/log"
+)
+
+// replicate a local thread on the hub
+func (s *synchronizer) restoreBucket(ctx context.Context, bucket string) error {
+
+	b, err := s.getBucket(ctx, bucket)
```
jsonsivar, line 12: Should this default to default bucket? Just wondering because if someone calls this on any bucket but `personal`, it will restore files from mirror path into that bucket :thinking:

core/textile/bucket_factory.go
```diff
@@ -408,13 +409,20 @@ func (tc *textileClient) ToggleBucketBackup(ctx context.Context, bucketSlug stri
 
 	if bucketSchema.Backup {
 		tc.sync.NotifyBucketBackupOn(bucketSlug)
+		tc.sync.NotifyBucketRestore(bucketSlug)
```
jsonsivar, line 412: Not fully clear on why we need to start a restore when backup is just toggled on?

jsonsivar (approved):
LGTM with some minor questions/comments.

core/textile/bucket_factory.go
maurycy: @jsonsivar 

Thank you for spotting this. Removed https://github.com/FleekHQ/space-daemon/pull/207/commits/2bcda3d1052a5da450071ec4301631a821e8b0ff#diff-904f54b080aa5edaf24e329308190d936166fff3aab8c0eaf9ff957b53dbf170L412

core/textile/sync/task-executors.go
maurycy: @jsonsivar 

Thank you for spotting this. Renamed in https://github.com/FleekHQ/space-daemon/pull/207/commits/2bcda3d1052a5da450071ec4301631a821e8b0ff#diff-26096467a39167c453134f075440f43a014ed45f4479be47d6ba72aaa7976829R27

core/textile/sync/synchronizer.go
maurycy: @jsonsivar 

Yes, indeed. Thank you for spotting this. Fixed in https://github.com/FleekHQ/space-daemon/pull/207/commits/2bcda3d1052a5da450071ec4301631a821e8b0ff#diff-47b13a650b4b80008ff5c0fde8630023ebe86568cc4ca509a299cd30b29cfa61R262

core/textile/sync/restore.go
maurycy: @jsonsivar 

Yes, fixed in https://github.com/FleekHQ/space-daemon/pull/207/commits/2bcda3d1052a5da450071ec4301631a821e8b0ff#diff-e4729a5c7a413a5b91b5b88b329f166525e3b3daec3b74e4991f7a19e1165c70R9 and, subsequently, in https://github.com/FleekHQ/space-daemon/pull/207/commits/8d0157394c577cfa40622fe5d94f4c7b11e8e90c#diff-e4729a5c7a413a5b91b5b88b329f166525e3b3daec3b74e4991f7a19e1165c70R10

core/textile/sync/restore.go
maurycy: @jsonsivar 

Yes, that's the meat now: https://github.com/FleekHQ/space-daemon/pull/207/files#diff-e4729a5c7a413a5b91b5b88b329f166525e3b3daec3b74e4991f7a19e1165c70R11

It needs very extensive testing but:

 * it relies only on the buckets now,
 * it uses the newly discovered `Item.Metadata.UpdatedAt`.

Ideally, the same logic should follow in https://github.com/FleekHQ/space-daemon/pull/207/files#diff-0fdaaca29057f19da1dabe9fdd5aa736707865612957999372f7ea54648e8993R218 but I'd wait for #207 

Also, I think that the newly added `bucket.Each()` could be used in many other places (we iterate over buckets everywhere) as well as `synchronizer.processBucketBackupOn()` should be smarter but, again, I'd wait for #207 and consider `bucket.Sync()` using the CID differences.

core/events/events.go
maurycy: @jsonsivar 

Thank you for spotting this, a left-over from reshuffles, fixed in d5279b68c5ed743ec1796130e8c29345266097d0

core/textile/sync/restore.go
maurycy: @dmerrill6 

Done, as per our sync in using `restoreFileTask` etc.

core/textile/sync/restore.go
maurycy: @jsonsivar 

I think it's a broader problem of allowing harmful operations...

grpc/handlers_backup.go
maurycy: @dmerrill6 

Keeping both :-) as in https://github.com/FleekHQ/space-daemon/pull/207/files#diff-c253e3f5d0b08f2e201f5184bf4e886c6d878c93d8483a74550a5fae2e18b972R218

core/textile/sync/pinning.go
maurycy: @dmerrill6 

Thank you for this suggestion, made it more generic in https://github.com/FleekHQ/space-daemon/pull/207/commits/ef0d6f4f30a11598e5339bfa00e382e3df1774a7#diff-26096467a39167c453134f075440f43a014ed45f4479be47d6ba72aaa7976829R27

core/textile/sync/restore.go
jsonsivar: Okay sounds good.  I would also triple check to make sure the `Metadata.UpdatedAt` is actually for the file and not the `roles` metadata. E.g., if someone updates the file but doesn't change the ACL would this get updated?

core/textile/sync/restore.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+
+			if localUpdatedAt >= mirrorUpdatedAt {
+				// do not overwrite: mirror is not newer
+				return nil
+			}
+		}
+
+		if s.eventNotifier != nil {
+			s.eventNotifier.SendFileEvent(events.NewFileEvent(itemPath, bucketSlug, events.FileRestoring, nil))
```
dmerrill6, line 50: It won't always be a new file event in case the file already existed on the local bucket but got an update from remote. We can also fix this in a following PR

core/textile/sync/synchronizer.go
```diff
@@ -135,6 +135,27 @@ func (s *synchronizer) NotifyBucketBackupOff(bucket string) {
...
+func (s *synchronizer) NotifyFileRestore(bucket, path string) {
+	t := newTask(restoreFileTask, []string{bucket, path})
+	s.enqueueTask(t, s.taskQueue)
+
+	s.notifySyncNeeded()
+}
+
+func (s *synchronizer) NotifyBucketStartup(bucket string) {
+	s.NotifyBucketRestore("personal")
+	s.NotifyBucketBackupOn("personal") // does nothing if !bucket.Backup
```
dmerrill6, line 154: Why do we use `"personal"` as a string here instead of `bucket` var?

core/textile/sync/task-executors.go
```diff
@@ -172,3 +182,48 @@ func (s *synchronizer) processBucketBackupOff(ctx context.Context, task *Task) e
...
+	bucket := task.Args[0]
+
+	if err := s.restoreBucket(ctx, bucket); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *synchronizer) processRestoreFile(ctx context.Context, task *Task) error {
```
dmerrill6, line 200: If we create a private function for the iterator at `restore.go` we could maybe call that function from here instead of repeating

dmerrill6 (approved):
LGTM, left some minor comments but looks good to merge and follow up with the minor corrections

PR #209, opened 2020-10-16
Shared file edit

dmerrill6:
# Change log

- Fix bug when updating a shared file that it didn't reflect on the local device.

- Now updating a shared file will update the remote version. NOTE: This is currently not notifying the other members of the shared file so they still see the old version. Created a card for fixing this (https://app.clubhouse.io/terminalsystems/story/19164/notify-all-members-when-a-shared-file-gets-updated)

- Updated buckets client to pull files from remote client, local cache and DHT (DHT not currently implemented). Whichever finishes first is returned and cached for the next time that it is requested. If we want to start pinning these files we can do it on this cache (uses IPFS client).

- A few refactors.

[ch18845]


core/sync/sync.go
```diff
@@ -265,7 +265,7 @@ func (bs *bucketSynchronizer) addFileInfoToStore(addFileInfo domain.AddWatchFile
 	if err := bs.store.SetString(getOpenFileKey(addFileInfo.LocalPath), string(out)); err != nil {
 		return err
 	}
-	reverseKey := getOpenFileReverseKey(addFileInfo.BucketKey, addFileInfo.BucketPath)
+	reverseKey := getOpenFileReverseKey(addFileInfo.BucketSlug, addFileInfo.BucketPath)
```
jsonsivar, line 268: Does `getOpenFileReverseKey` need to filter by a db ID as well? This was done on a previous PR so might have missed that but since we can have the same bucket name in different threads (and both my files and shared with me files will have `personal` as the name and for broader API usage this might be needed, but maybe could be post-alpha).

core/textile/secure_bucket_client.go
```diff
@@ -245,3 +268,220 @@ func (s *SecureBucketClient) decryptPathData(
...
+type pathPullingFn func(context.Context, string, string, io.Writer, ...bc.Option) (bool, error)
+
+type pullSuccessResponse struct {
+	file        *os.File
+	shouldCache bool
+}
+
+func (s *SecureBucketClient) racePullFile(ctx context.Context, key, encPath string, w io.Writer, opts ...bc.Option) error {
+	pullers := []pathPullingFn{s.pullFileFromLocal, s.pullFileFromClient, s.pullFileFromDHT}
+	// pullers := []pathPullingFn{s.pullFileFromLocal}
```
jsonsivar, line 281: This should be uncommented for now right?

jsonsivar (approved):
LGTM once the merge conflicts are fixed. Nice work. Also, as discussed on the call earlier we don't need to prioritize the card for notifying shared file edits since those will be remote right?

core/sync/sync.go
dmerrill6: Added `dbID` to the cache in a40be09

core/textile/secure_bucket_client.go
dmerrill6: Fixed in a40be09

dmerrill6:
> LGTM once the merge conflicts are fixed. Nice work. Also, as discussed on the call earlier we don't need to prioritize the card for notifying shared file edits since those will be remote right?

For shared files we now have a cache to store it once it's opened, so after getting the notification about a shared file being updated, we should simply clear this cache and it will be pulled from the hub the next time.

PR #212, opened 2020-10-19
Fix daemon initialization after deleting keys

dmerrill6:
[ch18895]


core/textile/model/model.go
```diff
@@ -112,7 +112,9 @@ func (m *model) findOrCreateMetaThreadID(ctx context.Context) (*thread.ID, error
 		return nil, err
 	}
 
-	if err := m.threads.NewDB(ctx, dbID, db.WithNewManagedThreadKey(managedKey)); err != nil {
+	err = m.threads.NewDB(ctx, dbID, db.WithNewManagedThreadKey(managedKey))
+	st := err.Error()
+	if err != nil && st != "rpc error: code = Unknown desc = db already exists" {
```
maurycy, line 117: @dmerrill6 

Is there a non-string we can use so we're future-proof?

maurycy (approved):
LGTM. 👍🏻

core/textile/model/model.go
dmerrill6: Not really :( It's internally calling a gRPC so the actual error object is not exposed

PR #213, opened 2020-10-20
Feature/ch19186/add shared link to shared with me

perfectmak:
**CHANGELOG**
- Update OpenSharedFile logic to add successfully opened file to ReceivedFiles thread
- Also, allows re-opening shared files without requiring password all the time
- Update the `SharedListDirectoryEntry` message proto schema to include `isPublicLink` to indicate the shared file is a public link item


core/textile/sharing.go
```diff
@@ -142,86 +183,131 @@ func (tc *textileClient) GetReceivedFiles(ctx context.Context, accepted bool, se
...
 		}
+		items = append(items, res)
+	}
 
-		ipfsHash := f.Item.Cid
-		name := f.Item.Name
-		isDir := false
-		size := f.GetItem().Size
-		ext := strings.Replace(filepath.Ext(name), ".", "", -1)
+	offset := files[len(files)-1].ID.String()
```
dmerrill6, line 200: Will this offset work even when we are merging two entity lists?

grpc/proto/space.proto
```diff
@@ -365,6 +365,7 @@ message SharedListDirectoryEntry {
   ListDirectoryEntry entry = 1;
   string dbId = 2;
   string bucket = 3;
+  bool isPublicLink = 4;
```
dmerrill6, line 368: Don't we also have to add the `password` field to the proto?

dmerrill6 (approved):
LGTM, left two questions

grpc/proto/space.proto
perfectmak: No, it is not useful. After the first time the open public file rpc is called, because that logic persists the password and automatically refetches it on subsequent calls.

So the frontend would pass an empty password when opening public link files from the shard with me response

core/textile/sharing.go
perfectmak: Yeah it should still work. We are not really merging two lists, it is only a single list being queried, but depending on the type of it's data we map it to a public shared file or a direct invite shared file.

grpc/proto/space.proto
dmerrill6: How do they pass the password then?

core/textile/sharing.go
dmerrill6: Understood 👍

grpc/proto/space.proto
perfectmak: They pass it only on the `OpenPublicFile` rpc request

dmerrill6: I don't see a field for password there

dmerrill6: I don't see a field for password there

core/space/services/services_sharing.go
```diff
@@ -217,7 +218,7 @@ func (s *Space) GenerateFilesSharingLink(ctx context.Context, encryptionPassword
 }
 
 // OpenSharedFile fetched the ipfs file and decrypts it with the key. Then returns the decrypted
-// files location.
+// files location. NOTE: This only opens public link shared files and not those shared via direct invites.
```
jsonsivar, line 221: Maybe we update the name to OpenPublicSharedFile?

core/textile/model/model.go
```diff
@@ -31,21 +31,31 @@ type Model interface {
...
 		ctx context.Context,
 		file domain.FullPath,
 		invitationId string,
 		accepted bool,
 		key []byte,
 	) (*ReceivedFileSchema, error)
+	CreateReceivedFileViaPublicLink(
+		ctx context.Context,
+		ipfsHash string,
+		fileKey string,
```
jsonsivar, line 44: Noticed we are passing in the pw here so maybe we can update the name (technically doesn't make a difference since they are both strings but just in case).

jsonsivar (approved):
LGTM once merge conflict is resolved and I think as per @dmerrill6 's comment, we need to add the password field for `OpenPublicFile` before merging.

PR #217, opened 2020-10-22
Feature/configure space site host

perfectmak:
**CHANGELOG**
- Enable configuring space storage site base url using the SPACE_STORAGE_SITE_URL environment variable
- Hotfix: using filename as path for public shared files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ build:
 	-X 'main.mongousr=${MONGO_USR}' \
 	-X 'main.mongopw=${MONGO_PW}' \
 	-X 'main.spaceapi=${SERVICES_API_URL}' \
+	-X 'main.spacestoragesiteurl=${SPACE_STORAGE_SITE_URL}' \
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -7,6 +7,7 @@ build:
 	-X 'main.mongousr=${MONGO_USR}' \
 	-X 'main.mongopw=${MONGO_PW}' \
 	-X 'main.spaceapi=${SERVICES_API_URL}' \
+	-X 'main.spacestoragesiteurl=${SPACE_STORAGE_SITE_URL}' \
```
jsonsivar, line 10: I think we also need this added to the github actions files (.github/workflows/release.yml and .goreleaser.yml) (and also PRD vs DEV if they will be diff).

maurycy (approved):
LGTM. 🚀

Makefile
perfectmak: @jsonsivar Thanks for pointing that out. I've added it to the release and goreleaser yaml files.

PR #219, opened 2020-10-23
Delete account v1

jsonsivar:
### Change Log
Extended @studna's work to delete content (note: this is an initial v but we want to later do proper shut downs and clearing of data dirs across the board but that's going to require some refactoring at very core levels).

### Todo
* Add mocked unit tests


app/app.go
```diff
@@ -93,12 +93,13 @@ func (a *App) Start(ctx context.Context) error {
...
 	a.Run("FolderWatcher", watcher)
 
+	var ipfsNode *node.IpfsNode
 	// setup local ipfs node if Ipfsnode is set
 	if a.cfg.GetBool(config.Ipfsnode, true) {
 		// setup local ipfs node
-		node := node.NewIpsNode(a.cfg)
-		err = a.RunAsync("IpfsNode", node, func() error {
-			return node.Start(ctx)
+		ipfsNode = node.NewIpsNode(a.cfg)
```
perfectmak, line 100: I think it would be better to do this:
```suggestion
		ipfsNode := node.NewIpsNode(a.cfg)
```

and delete the 
```
var ipfsNode *node.IpfsNode
```

core/space/services/services.go
```diff
@@ -7,9 +7,11 @@ import (
 
 	"github.com/FleekHQ/space-daemon/core/textile/hub"
 	"github.com/FleekHQ/space-daemon/core/vault"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/FleekHQ/space-daemon/config"
 	"github.com/FleekHQ/space-daemon/core/env"
+	node "github.com/FleekHQ/space-daemon/core/ipfs/node"
```
perfectmak, line 14:
```suggestion
 "github.com/FleekHQ/space-daemon/core/ipfs/node"
```
The extra `node` at the beginning is redundant, you can remove it

maurycy (approved):

core/keychain/keychain.go
```diff
@@ -222,15 +223,17 @@ func (kc *keychain) DeleteKeypair() error {
 
 	// Note: currently ignoring error on keychain removal because it's failing randomly.
 	// Use GenerateKeyPair with override option instead.
-	ring.Remove(PrivateKeyStoreKey)
+	err = ring.Remove(PrivateKeyStoreKey)
+	if err != nil {
+		log.Error("Error removing keychaing from keyring", err)
```
perfectmak, line 228: Maybe this should be a `log.Warn()` instead.

dmerrill6 (approved):
LGTM

core/space/services/services.go
jsonsivar: That's not compiling. When I removed the "node" at the front, it complains on `*node.IpfsNode` that node is undefined.  Is that because `node` itself is not a package? I.e., it's package name is still `ipfs`.  Should it be it's own package? @maurycy  Not sure if I'm missing some go mod detail.

core/keychain/keychain.go
jsonsivar: This is a weird one.  It's actually an error scenario hence the `Error` but we are just ignoring it temporarily because it happens intermittently. Maybe it's something that's fixed cc @dmerrill6. Either way, I'm removing the whole `if` block in the meantime to leave it like how it was.

app/app.go
jsonsivar: Fixed in [9bc2f71](https://github.com/FleekHQ/space-daemon/pull/219/commits/9bc2f71364d4b739cf4172e80aa9ef7f3d7e3885).

jsonsivar:
@dmerrill6 @maurycy @perfectmak can you kindly review this again? The change most impactful change is here https://github.com/FleekHQ/space-daemon/pull/219/commits/68e8f40edb32e38d1f3ce9a037d445a89165cb28 and spoke with @dmerrill6 that moving the sync init to where it is now is not the best, so we can decide if that's okay for now or we should tweak it in this PR.

core/ipfs/node/node.go
```diff
@@ -167,6 +159,13 @@ func (node *IpfsNode) start() error {
 }
 
 func (node *IpfsNode) stop() error {
+	node.IsRunning = false
+
+	err := node.stop()
```
perfectmak, line 164: QQ: Wouldn't this cause an infinite loop? `node.stop()` calling itself with no stop condition?

perfectmak (approved):
@jsonsivar LGTM!. I left a question though that should be checked

core/ipfs/node/node.go
jsonsivar: Oh thanks good catch.  Meant to call it on the inner node object. Fixed in [d4ffaef](https://github.com/FleekHQ/space-daemon/pull/219/commits/d4ffaef036f1f57511eeeaa7a53b414e7c2dea8f).

PR #220, opened 2020-10-23
Listen for textile client updates

dmerrill6:
# Change log

Now when we detect a change in a remote bucket, we trigger the restoration flow.

## Notes

Haven't been able to fully make it work yet. The restoration flow is not listing remote files but instead only listing local ones, so the remote changes aren't synced over (testing the app by running the same account in two computers). Wanted to push this anyway in case I can get more eyes on it.

Also another TODO is listening for updates in `shared_with_me` thread to clear out the cache, but want to fix the previous point before.


core/textile/event_handler.go
```diff
@@ -27,3 +28,28 @@ func (h *defaultListenerHandler) OnRemove(bucketData *bucket.BucketData, listenE
...
+	synchronizer sync.Synchronizer
+}
+
+func newRestorerListenerHandler(synchronizer sync.Synchronizer) *restorerListenerHandler {
+	return &restorerListenerHandler{
+		synchronizer: synchronizer,
+	}
+}
+
+func (h *restorerListenerHandler) OnCreate(bucketData *bucket.BucketData, listenEvent *tc.ListenEvent) {
```
maurycy, line 42: @dmerrill6 

Do you think it's possible to make it more granular in the next release?

maurycy (approved):
LGTM. 👍🏻

core/textile/event_handler.go
dmerrill6: I think it's possible, although I think most of the events trigger OnSave (think OnCreate and OnDelete are triggered on bucket creation/deletion and not at file level)

PR #221, opened 2020-10-26
Feature: Files Search Schema and stub implementation [ch19245]

perfectmak:
**CHANGELOG**
- Create schema for searching files through a `SearchFiles` rpc method.
- Currently, the default search implementation just returns list of file entries in the root path of the current users default bucket. This implementation would be fixed in the upcoming PR's.
- Refactors the logic to convert a domain `FileInfo` into `ListDirectoryEntry`. To reduce duplicate code of the same logic


grpc/proto/space.proto
```diff
@@ -308,6 +308,23 @@ service SpaceApi {
       body: "*"
     };
   }
+
+  // Search for files across all users bucket
+  rpc SearchFiles(SearchFilesRequest) returns (SearchFilesResponse) {
+    option (google.api.http) = {
+      post: "/v1/search/files"
```
jsonsivar, line 315: Think this should be `get`

gpuente (approved):
LGTM

jsonsivar (commented):
LGTM just needs to be a `GET`.

grpc/proto/space.proto
perfectmak: True.... I've fixed it.

maurycy (approved):
LGTM. 🚀

PR #226, opened 2020-10-30
Replication and restoration of Metathread

dmerrill6:
# Change log

- Metathread is now correctly replicated and then restored when signing in.

- Recreate threads and buckets when needed.

- Fixes on file restore task to not error out when the pulled file does not exist on the local bucket.

[ch19344]

TODO: Test file syncing on edit for shared files.


maurycy (approved):

core/textile/utils/utils.go
```diff
@@ -205,15 +228,63 @@ func FindOrCreateDeterministicThreadID(ctx context.Context, variant Deterministi
...
+
+	hubCtx, err := hubAuth.GetHubContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// i, err := netc.GetThread(ctx, dbID)
+	_, err = hnetc.GetThread(hubCtx, dbID)
+	replThreadExists := err == nil
+	if !replThreadExists && shouldForceRestore {
```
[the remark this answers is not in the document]
maurycy: @perfectmak 

Could you take a look?

core/textile/model/model.go
```diff
@@ -63,49 +64,22 @@ type Model interface {
...
-		return &threadID, nil
-	}
-
-	if _, err := m.netc.AddReplicator(ctx, threadID, hubma); err != nil {
-		log.Error("error while replicating metathread", err)
-		// Not returning error in case the user is offline (it should still work using local threads)
-	}
-
-	return &threadID, nil
+	return utils.FindOrCreateDeterministicThread(ctx, utils.MetathreadThreadVariant, metaThreadName, m.kc, m.st, m.threads, m.ht, m.cfg, m.netc, m.hnetc, m.hubAuth, m.shouldForceRestore, GetAllCollectionConfigs())
```
perfectmak, line 82: Long argument lists like this one should be split into multiple lines

perfectmak (approved):

core/textile/utils/utils.go
```diff
@@ -205,15 +228,63 @@ func FindOrCreateDeterministicThreadID(ctx context.Context, variant Deterministi
...
+	hubmaWithThreadID := hubmaStr + "/thread/" + dbID.String()
+
+	hubCtx, err := hubAuth.GetHubContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// i, err := netc.GetThread(ctx, dbID)
+	_, err = hnetc.GetThread(hubCtx, dbID)
+	replThreadExists := err == nil
```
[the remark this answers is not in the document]
dmerrill6: According to Sander this should be enough (i.e. if the thread exists then err will be nil)

dmerrill6:
Added restoration of each bucket thread, requesting some more eyes before merging 🙏

core/textile/bucket_factory.go
```diff
@@ -452,3 +465,93 @@ func GetDefaultBucketSlug() string {
...
+	}
+
+	threadsInitialized := true
+	for _, b := range bucketList {
+		dbID, err := utils.ParseDbIDFromString(b.DbID)
+		if err != nil {
+			return err
+		}
+
+		if _, ok := dbs[*dbID]; !ok {
```
maurycy, line 493: @dmerrill6 

I hope there is no desync between `model` and `ListDBs` :-)

core/textile/bucket_factory.go
```diff
@@ -452,3 +465,93 @@ func GetDefaultBucketSlug() string {
...
+
+// Attempts to restore buckets from a hub replication
+// Returns nil if there's nothing to restore or the restoration succeeded
+func (tc *textileClient) restoreBuckets(ctx context.Context) error {
+	bucketList, err := tc.GetModel().ListBuckets(ctx)
+	if err != nil {
+		return err
+	}
+
+	if len(bucketList) == 0 && tc.shouldForceRestore {
```
maurycy, line 477: @dmerrill6 

Is `tc.shouldForceRestore` check necessary here?

maurycy (approved):
LGTM. 🚀

core/textile/bucket_factory.go
dmerrill6: It's here because I don't want to error out when creating a new account (as it has no buckets), only when restoring an existing account

PR #231, opened 2020-11-03
gprc stub: torus

maurycy:


studna:
I think it depends if we want to handle encrypt/decrypt on daemon side or let FE to encrypt/decrypt textile keys (then it’d be more like `ExportTextileKeys` and `ImportTextileKeys`. I think handling encryption on FE makes more sense, so we don’t bundle it into daemon. 

I mean we provide daemon as open source, this could be nice way how to allow developers to handle backup/restore on their own (someone can use AppSync + Cognito, others can use Firebase Auth and so on). Otherwise we force to use ETH (ECDSA) key pairs and encryption.

jsonsivar:
> I think it depends if we want to handle encrypt/decrypt on daemon side or let FE to encrypt/decrypt textile keys (then it’d be more like `ExportTextileKeys` and `ImportTextileKeys`. I think handling encryption on FE makes more sense, so we don’t bundle it into daemon.
> 
> I mean we provide daemon as open source, this could be nice way how to allow developers to handle backup/restore on their own (someone can use AppSync + Cognito, others can use Firebase Auth and so on). Otherwise we force to use ETH (ECDSA) key pairs and encryption.

Yeah, so in the card we updated it to be 2 functions specifically: encrypt the key for the FE so they can upload, and import a key which the FE will pass.We can still handle encryption on daemon side but make the functions more adapted to this flow. I think the restore one is missing altogether.

studna:
> Yeah, so in the card we updated it to be 2 functions specifically: encrypt the key for the FE so they can upload, and import a key which the FE will pass.We can still handle encryption on daemon side but make the functions more adapted to this flow. I think the restore one is missing altogether.
I understand that, just mentioning it'd be more flexible if daemon export/import raw textile keys, so any app can decide how to handle encryption/decryption. I mean in this PR we allow to provide ETH/ECDSA compatible key, but the algorithm itself would be hardcoded on the daemon (so you can't use different crypto algorithms).

PR #233, opened 2020-11-03
Fix/edit shared files

dmerrill6:
# Change log

- Make cache based off CID so that when we restore a file after detecting a remote change, open file will fetch from the bucket instead of the cache for the new file.

[ch18845]


jsonsivar (approved):

maurycy (approved):
LGTM. 🚀

PR #237, opened 2020-11-05
Hotfix: call to disable sync and make sure its called on delete account

jsonsivar:


maurycy (approved):
@jsonsivar 

I think it's good to go. Could you confirm, @dmerrill6 ?

dmerrill6 (approved):

PR #238, opened 2020-11-05
Release to master

jsonsivar:
### Features
* Partial text search

### Bug Fixes
* Properly map file info to dir entry for file events
* Remove file info stream and add bucket/dbid to file events
* Search init error (normalize user dir and path string)


maurycy (approved):
LGTM. 🚀

PR #239, opened 2020-11-05
sqlite: use usr.HomeDir

maurycy:
The pull request fixes the following error:

```
ERRO[0169] Error initializing files search index -- ERROR -- failed to open database: unable to open database file: no such file or directory[]
```

The tilde is replaced in `Start()` instead of `NewSearchEngine()` because the latter returns only `*sqliteFilesSearchEngine`, without the `err`, and the former is called only once anyway.

### DIFF
--- a/config/map_config.go
+++ b/config/map_config.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"os"
+	"os/user"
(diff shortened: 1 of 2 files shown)


jsonsivar (approved):
LGTM

PR #240, opened 2020-11-06
Enable CGO for windows release [ch19662]

perfectmak:


jsonsivar (approved):
LGTM

maurycy (approved):
@perfectmak 

LGTM, but... @dmerrill6 do you remember the reasoning here, or the original issue never occured on Windows? 7f60a3d4

dmerrill6:
> @perfectmak 
> 
> 
> 
> LGTM, but... @dmerrill6 do you remember the reasoning here, or the original issue never occured on Windows? 7f60a3d4

CGO uncommented was giving issues with the keyring. But if this was enabled it wouldn't compile for me because goreleaser wasn't compatible with cgo. If this compiles on windows I don't think there would be issues enabling it.

PR #241, opened 2020-11-06
Hotfix: add type for passphrase backup

jsonsivar:
This is to enable a single user working with both password backup as well as Torus.  It adds a type field to the storage of the backup.


maurycy (approved):
LGTM. 🚀

dmerrill6 (approved):

PR #243, opened 2020-11-09
Feature: Replace Sqlite Search with Bleve

perfectmak:
**CHANGELOG**
- Change default files search engine from sqlite to bleve

Bleve is built natively in golang and provides full text search capabilities, so it would serve our purpose well.
I'm replacing the sqlite implementation so we don't have to deal with setting up cgo in our deployment/development process and also prevent windows releases from being tedious.


dmerrill6 (approved):
LGTM

maurycy (approved):
LGTM. 🚀

PR #246, opened 2020-11-09
Security: App tokens

dmerrill6:
# Change log

- Now all endpoints require the client to send an App Token, otherwise the request gets rejected. 

- Add `InitializeMasterAppToken`, which returns a token with full access only if there's no other master token already created.

- Add tests

- Add a stub for `GenerateAppToken`. Didn't implement it yet (but the logic is there) since it will require some kind of "app store" view on the client so that users can selectively authorize apps with scoped access (e.g. Do you authorize X to ListDir and OpenFile?).

Tested it on the client-side in this PR https://github.com/FleekHQ/space-client/pull/102


core/permissions/app_token.go
```diff
@@ -0,0 +1,96 @@
...
+const tokenSecretLength = 30
+
+type AppToken struct {
+	Key         string
+	Secret      string
+	IsMaster    bool
+	Permissions []string
+}
+
+// Token structure is [KEY].[SECRET].[IS_ADMIN].[PERMISSION1]_[PERMISSION2]...
```
perfectmak, line 22: Perhaps marshalling into a json string might be better and less error prone? Is there a reason why this format was chosen?

perfectmak (commented):
LGTM... I left some questions.

core/keychain/app_token.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		if err := ring.Set(keyring.Item{
+			Key:   getMasterTokenStKey(),
+			Data:  []byte(permissions.MarshalFullToken(tok)),
+			Label: "Space App - Master App Token",
+		}); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
dmerrill6: Yea, wanted to access both master and regular tokens with the same logic, but at the same time needed some way to know if there was a master token present to prevent overwrites.

core/keychain/app_token.go
```diff
@@ -0,0 +1,75 @@
...
+	// Prevents overriding even if user logged out and logged back in (which clears the store)
+	_, err = ring.Get(getMasterTokenStKey())
+	if err == nil && tok.IsMaster {
+		return ErrMasterTokenAlreadyExists
+	}
+
+	err = ring.Set(keyring.Item{
+		Key:   AppTokenStoreKey + "_" + tok.Key,
+		Data:  []byte(permissions.MarshalFullToken(tok)),
+		Label: "Space App - App Token",
```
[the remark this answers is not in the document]
dmerrill6: I don't really think it's necessary as this is just so that it gets displayed like that when you browse them in the keychain. It's not like the key to access the data later on.

core/permissions/app_token.go
dmerrill6: I can't recall why I went for this approach but you're right, I'll update it to a json marshal.

dmerrill6: Fixed in 9fe539a

PR #248, opened 2020-11-11
GetSharedByMeFiles()

maurycy:
[ch19726]


core/textile/model/sent_file.go
```diff
@@ -0,0 +1,177 @@
...
+	if rawFiles == nil {
+		return []*SentFileSchema{}, nil
+	}
+
+	files := rawFiles.([]*SentFileSchema)
+	return files, nil
+}
+
+// XXX: this is to reuse the builders in the sharing.go
+func (sf *SentFileSchema) ReceivedFileSchema() *ReceivedFileSchema {
```
dmerrill6, line 144: Please call this in `GetAllCollectionConfigs` in model.go so that it gets restored too.

dmerrill6 (approved):
Other than the comment LGTM

core/textile/model/sent_file.go
maurycy: @dmerrill6 

Thank you for spotting this!

PR #249, opened 2020-11-11
Fix: Speed up public link generation

perfectmak:
**CHANGELOG**
- Setup tracing with jaegar and collect trace information on the `GeneratePublicFileLink` rpc
- Create default public share bucket on textile client initialization to prevent initialization during first time of sharing (should at least speed up some)

**Details of issue**
Uploads are still slow and tracing (see image below) shows that it take a long time while uploading to the bucket. The example in the image was a 3.5MB file and it took ~7mins to upload.
 - It looks like uploads on the hub are actually quite slow (and I don't think it is only public bucket).

Screenshot 2020-11-11 at 13 27 52

### DIFF
--- a/core/space/services/services_sharing.go
+++ b/core/space/services/services_sharing.go
@@ -13,6 +13,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/opentracing/opentracing-go"
+
 	"github.com/FleekHQ/space-daemon/config"
 
 	"github.com/FleekHQ/space-daemon/log"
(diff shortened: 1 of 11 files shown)


core/space/services/services_sharing.go
```diff
@@ -46,56 +51,68 @@ func (s *Space) GenerateFileSharingLink(
 		return domain.FileSharingInfo{}, err
 	}
 
+	encryptedFile, err := s.encryptBucketFile(ctx, encryptionPassword, path, bucket)
+	if err != nil {
+		return EmptyFileSharingInfo, errors.Wrap(err, "file encryption failed")
+	}
+
+	log.Printf("Copy successful")
```
dmerrill6, line 59: This maybe can use `log.Debug`

dmerrill6 (approved):
LGTM

PR #250, opened 2020-11-11
Feature: Improve infix search on files search

perfectmak:
Reference: [ch19766]

**CHANGELOG**
- Update files search query to perform wildcard (prefix/suffix) search on files and folders.
- Setup proper mapping for index to ignore noisy fields when searching for records (to avoid irrelevant results).


dmerrill6 (approved):
LGTM

PR #259, opened 2020-11-18
Fix: Finalize Fuse Installation and Testing

perfectmak:
**CHANGELOG**
- Properly handles fuse status/state to make it easier to client to determine what action to perform. States are either `UNSUPPORTED` if on an operating system without fuse support, `NOT_INSTALLED` when fuse is not detected, `UNMOUNTED` when fuse is installed but the Space drive is not mounted and `MOUNTED` for when the drive is mounted.
- Fixes some bugs with handling opened files in the drive


app/app.go
```diff
@@ -103,7 +105,7 @@ func (a *App) Start(ctx context.Context) error {
 			return node.Start(ctx)
 		})
 		if err != nil {
-			log.Error("Error starting embedded IPFS node", err)
+			log.Error("ERROR starting embedded IPFS node", err)
```
perfectmak, line 108: This `Error` -> `ERROR` changes are unintended and I'll fix them before merging

perfectmak:
Hi guys, I'll want to merge this soon. So the updated API can be used for integrations.

PR #260, opened 2020-11-20
Hotfix: add type for recover, and some cleanups

jsonsivar:
Added the `type` field to the recover method and also updated the enum to be Google, Twitter, etc instead of Eth.


grpc/proto/space.proto
```diff
@@ -584,6 +584,7 @@ message BackupKeysByPassphraseResponse {}
 message RecoverKeysByPassphraseRequest {
   string uuid = 1;
   string passphrase = 2;
+  KeyBackupType type = 3;
```
perfectmak, line 587: You should also update the `enum KeyBackupType` definition of the proto file to contain the new fields

perfectmak (changes requested):
Generally looks good. I requested a change that you missed in the proto defintions.

maurycy (approved):
LGTM. 🚀

grpc/proto/space.proto
jsonsivar: Good catch, fixed in [8f5fafc](https://github.com/FleekHQ/space-daemon/pull/260/commits/8f5fafc86fe33d2152906df537be2b92ef54bf80)

perfectmak (approved):
LGTM. There are some conflicts though.

PR #261, opened 2020-11-23
Fix: Re-mounting fuse failure

perfectmak:
**CHANGELOG**
- Expose mountPath in FuseDriveResponse. This should allow the client to have access to the directory.
- Fix an issue where fuse drive is unable to re-mount if the daemon is abruptly stopped.


dmerrill6 (approved):
LGTM

PR #263, opened 2020-11-24
Add RemoveDirOrFile grpc method

dmerrill6:
[ch20270]


core/space/services/services_fs.go
```diff
@@ -703,3 +703,24 @@ func (s *Space) addFile(ctx context.Context, sourcePath string, targetPath strin
 		Bytes:      fileSize,
 	}, err
 }
+
+// Removes a file or directory from a bucket
+// Note: If removing a file a user has been shared, it should remove themselves as a member instead of removing the actual file.
```
perfectmak, line 708: > it should remove themselves as a member instead of removing the actual file.

@dmerrill6 Are you still going to work on this? I don't think `b.DeleteDirOrFile()` does this

dmerrill6: Oh it was badly explained. I meant there should be another method to do that call.
I'll update that comment

perfectmak (approved):

maurycy (approved):

PR #269, opened 2020-12-01
Feature: Remove users access from a shared file

perfectmak:
**CHANGELOG**
- Adds a rpc method UnshareFilesViaPublicKey` that revokes access of a user from a shared file.
- Unsharing a file sends a revoked invitation notification to the user and when the user receives this notification, it should verify and remove from them from his list of received files.


jsonsivar (approved):
LGTM

maurycy (approved):
LGTM. 🚀

PR #270, opened 2020-12-01
services_fs: tmp files only in tmp dir

maurycy:
ref https://github.com/FleekHQ/space-daemon/pull/267#discussion_r532205241


perfectmak (approved):
LGTM.

PR #271, opened 2020-12-01
ipfs: use dht

maurycy:
[ch20181]


core/space/services/services_sharing.go
```diff
@@ -271,7 +271,7 @@ func (s *Space) OpenSharedFile(ctx context.Context, hash, password, filename str
 		}
 	}
 
-	encryptedFile, err := s.tc.DownloadPublicGatewayItem(ctx, parsedCid)
+	encryptedFile, err := s.tc.DownloadPublicItem(ctx, parsedCid)
```
jsonsivar, line 274: Tagging @perfectmak here to make sure this will work.  Because it will expect the CID for the public file to be available via DHT locally. Maybe what we should do is a "race pull" similar to what we need to do for `secureBucketClient.PullPath`.

jsonsivar (approved):
LGTM

core/space/services/services_sharing.go
perfectmak: Yes @maurycy, this `OpenSharedFile()` function needs to download the item from the public hub gateway. We need to test if this new DHT download will be able to fetch from the public hub gateway.

perfectmak: So yeah, maybe leave the `DownloadPublicGatewayItem` as it is because it performs a different function from download via the local node.

studna:
Just idea based on brief discussion with @jsonsivar : it might be more efficient if we use https://docs.ipfs.io/reference/http/api/#api-v0-get with `output` argument, so there's no need to buffer output on HTTP layer.

PR #276, opened 2020-12-03
secure_bucket_client: do not send to a closed chan

maurycy:


perfectmak (commented):
Honestly, I'm not sure what is going on here , so I'll leave it to Janison to take a look.

